PR #21, opened 2018-05-30
Parametrize page navigation timeout (fixes #20)

### LINKED ISSUE #20: Page navigation timeout is not parametrized
Hello there :wave: 

thank you for this cool package in advance, i'm using it to access information from sites which do not offer APIs and it is a very helpful tool.

On of the things i require to be able is to set the timeout when calling `\HeadlessChromium\Page::navigate`.
Currently there is no parameter, therefor `\HeadlessChromium\Communication\Connection::$sendSyncDefaultTimeout` will be used which can't be set and is hardcoded to 3 seconds. When downloading files from some pages the download might take longer than 3 seconds (because the file is big/needs to be generated by the server first) and `$page->navigate($downloadUrl);` will throw an exception.

I will provide a non-breaking pull request shortly and i'd be glad if you could merge it into your awesome lib :slightly_smiling_face:

vertexvaar:


gsouf:
Hi @vertexvaar as I said in the related issue, I proposed a more general fix, so that you only have to deal with a single timeout.

Btw, thanks for your contribution, that's appreciated.

PR #34, opened 2018-08-29
Allow chrome headless to run inside a docker container

tanasecosminromeo:
Add 'nosandbox' option that will allow running chrome headless inside a docker container. Find socket uri regex updated as well as the format is a bit different when running in docker.

Other notes: I've used `debian:stretch-slim` to run successfully chrome inside the container. Had some issues with `alpine` as the socket was TCP, not a web socket. Will investigate further and try and make it compatible with this as well.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -28,6 +28,7 @@ class Browser
     {
         $this->connection = $connection;
 
+        // listen for target created
         $this->connection->on(Connection::EVENT_TARGET_CREATED, function (array $params) {
 
             // create a session for the target
@@ -40,8 +41,21 @@ class Browser
             );
         });
 
+        // listen for target info changed
+        $this->connection->on(Connection::EVENT_TARGET_INFO_CHANGED, function (array $params) {
+
+            // get target by id
+            $target = $this->getTarget($params['targetInfo']['targetId']);
+
+            if ($target) {
+                $target->targetInfoChanged($params['targetInfo']);
+            }
+        });
+
+        // listen for target destroyed
         $this->connection->on(Connection::EVENT_TARGET_DESTROYED, function (array $params) {
 
+            // get target by id
             $target = $this->getTarget($params['targetId']);
 
             if ($target) {
@@ -108,6 +122,9 @@ class Browser
 
         // Page.enable
         $page->getSession()->sendMessageSync(new Message('Page.enable'));
+
+        // Network.enable
+        $page->getSession()->sendMessageSync(new Message('Network.enable'));
 
         // Page.setLifecycleEventsEnabled
         $page->getSession()->sendMessageSync(new Message('Page.setLifecycleEventsEnabled', ['enabled' => true]));
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -78,7 +78,7 @@ class Browser
      * @throws CommunicationException
      * @return Page
      */
-    public function createPage(): Page
+    public function createPage(array $options = []): Page
     {
 
         // page url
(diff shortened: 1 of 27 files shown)


src/Browser.php
```diff
@@ -78,7 +78,7 @@ public function close()
      * @throws CommunicationException
      * @return Page
      */
-    public function createPage(): Page
+    public function createPage(array $options = []): Page
```
tanasecosminromeo: **Useful Example**
`$browser->createPage(['preScript' => file_get_contents('do-not-detect.js')]);`

*do-not-detect.js*
```
// overwrite the `languages` property to use a custom getter
Object.defineProperty(navigator, 'languages', {
    get: function() {
        return ['en-US', 'en'];
    },
});

// overwrite the `plugins` property to use a custom getter
Object.defineProperty(navigator, 'plugins', {
    get: function() {
        // this just needs to have `length > 0`, but we could mock the plugins too
        return [1, 2, 3, 4, 5];
    },
});

// overwrite the `plugins` property to use a custom getter
Object.defineProperty(window, 'chrome', {
    get: function() {
        // this just needs to have `length > 0`, but we could mock the plugins too
        return [1, 2, 3, 4, 5];
    },
});

// set navigator.webdriver to false
Object.defineProperty(navigator, 'webdriver', {
    get: () => false,
});


// Simulate navigator permissions;
const originalQuery = window.navigator.permissions.query;
window.navigator.permissions.query = (parameters) => (
    parameters.name === 'notifications' ?
        Promise.resolve({ state: Notification.permission }) :
        originalQuery(parameters)
);
```

gsouf:
Hi @tanasecosminromeo 

I will review your changes and merge them once it's done.

tanasecosminromeo:
@gsouf Yep, sorry for adding more than one change in the same PR

- 1st one is to allow using chrome-php with docker
- 2nd one is a fix for the user agent, if the parameter is unescaped bash will give an error, and people may not think to do that when they send the param in `browserCreate`
- 3nd one is to allow including scripts before pageload. Useful to add libraries before page load, and you can also use it to try and make chrome headless undetectable - I followed the stuff here `https://antoinevastel.com/bot%20detection/2018/01/17/detect-chrome-headless-v2.html`

I've also done printToPdf, but currently only implemented landscape and printBackground options. When I have a bit of time will implement all options and do another PR.

Another thing I'm thinking is to create some sort of caching for createBrowser and pagecreate and to keep a chrome DevTools session active that the script can reconnect to. So pretty much to only create the browser if it's not already created, but we need this in a smart way that will cache the option and recreate if needed + add some sort of kill for old browser sessions. Happy to submit more PR once I do some changes tho :)

gsouf:
Hi @tanasecosminromeo 

Thanks for your work. I'll be doing a few minor changes and do other work before I publish the release.

As for caching, I dont think it's good to implement this in the library, the goal of the library is to provide an unambiguous way to start new instances of chrome and to create web pages. I think it's better to do it in some wrapper which the only goal is to reuse pages and existing browser.

gsouf:
@tanasecosminromeo I have been doing a few modifications in your implementation of preScript feature.

The way you implemented it the same script was added once again everytime we navigated to the page. The right way to use it will be:

```php
$page->addPreScript('some script');
```

or you can add it directly on the browser:

```php
$browser->setPagePreScript('some script');
$p1 = $browser->createPage();
$p2 = $browser->createPage();

// both of $p1 and $p2 have the preScript added

$p1->addPreScript('some other script');

// $p1 has the 2 scripts enabled

$browser->setPagePreScript(null);
$p3 = $browser->createPage();

// $p3 has no prescript

```

gsouf:
Also note that I renamed option ``nosandbox`` to ``noSandbox``

PR #45, opened 2018-10-25
Test for getLayoutMetrics with larger content

kaero598:


gsouf:
I did it already, so I'm closing the PR

PR #48, opened 2018-11-12
Add Timeout to getReturn functions

chris-ware:
Allow timeouts to be set on getReturn functions


gsouf:
Hi @chris-ware 

Thanks for the contribution. Can you do a minor change and use "int" as type hint for timeouts? Ie: ``public function waitForResponse(int $timeout = null)``.

gsouf:
@chris-ware  thanks for the changes. I'll tag it tomorrow with an other change

gsouf:
@chris-ware published in version ``0.6.0``

PR #51, opened 2018-11-16
pdf options

marios88:
added pdf options:  
displayHeaderFooter  
preferCSSPageSize  
marginTop  
marginBottom  
marginLeft  
marginRight  

Ref: https://github.com/chrome-php/headless-chromium-php/issues/49


gsouf:
Hi @marios88 thanks for the PR. I'll add it with next version.

PR #58, opened 2018-12-06
Adaptation for php56

Kozin-br:


Kozin-br:
i thint it is no so well commit

PR #80, opened 2019-03-22
missing Header & Footer options + fixed comments

MUGagency:


gsouf:
Thank you @MUGagency sorry for delay!

PR #95, opened 2019-06-14
Handle operation timeout when closing all pages.

mschwerin:


mschwerin:
Just added a solution for non-escaped command line parameters.

src/BrowserFactory.php
```diff
@@ -16,22 +16,22 @@
...
     {
-        // auto guess chrome binaries
-        if (null === $chromeBinaries) {
+        // auto guess chrome binary
+        if (null === $chromeBinary) {
             $envChromePath = getenv('CHROME_PATH');
 
             if ($envChromePath) {
-                $chromeBinaries = '"' . $envChromePath . '"';
+                $chromeBinary = '"' . $envChromePath . '"';
```
jfcherng: We don't need to add the quote here then. It breaks the following currently working code.

```php
putenv('CHROME_PATH=C:\\Program Files (x86)\\Google\\Chrome\\Application\\chrome.exe');
```

src/Browser/BrowserProcess.php
```diff
@@ -111,16 +111,13 @@ public function start($binaries, $options)
...
         // get args for command line
-        $args = $this->getArgsFromOptions($options);
-
-        // process string
-        $processString = $binaries . ' ' . implode(' ', $args);
+        $args = $this->getArgsFromOptions($binary, $options);
 
         // log
-        $this->logger->debug('process: starting process: ' . $processString);
+        $this->logger->debug('process: starting process: ' . implode(' ', $args));
```
jfcherng: This should be
```php
$this->logger->debug('process: starting process: ' . $process->getCommandLine());
```
and be put after `$process` has been initiated before calling `$process->start()`.

gsouf:
hi @mschwerin appreciate the work but next time try to break different features into different PR. It's hard to review when it mixes multiple concerns.

Does that work with `symfony/process 3.x`?

PR #103, opened 2019-09-05
Add timeout for getBase64

TimIgoe:
Needed to add a timeout to getBase64 for large images, otherwise it was taking too long to capture the screenshot and return it to the caller


gsouf:
Thanks @TimIgoe

PR #111, opened 2019-09-30
Fix Symfony deprecation

enumag:
```
Passing a command as string when creating a "\Symfony\Component\Process\Process"
instance is deprecated since Symfony 4.2, pass it as an array of its arguments
instead, or use the "Process::fromShellCommandline()" constructor if you need
features provided by the shell.
```


gsouf:
Hi @enumag it's been fixed by other work, but thanks for your work, much appreciated!

PR #133, opened 2020-01-27
Close browser method

Alexsisukin:
I need a method for closing the browser, I work through the ``connectToBrowser`` method, I got an error, it seems to me that the 63 version is no longer relevant, you can refuse it


gsouf:
Hi @Alexsisukin  thanks for your PR merging now

PR #137, opened 2020-02-21
V0.9.x dev

ttk:
- Fixed a couple of failing unit tests (Windows)
- added some additional features with unit tests

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -191,4 +191,19 @@ class Browser
     {
         return array_values($this->targets);
     }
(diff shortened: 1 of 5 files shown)


ttk:
I didn't write a unit test for setDownloadPath().  For me, it sometime didn't work (and therefore I'm not actually using it in my project).  The Chrome devs are apparently going to replace Page.setDownloadBehavior with something else....so if you don't want to merge that feature, that's cool with me.

PR #163, opened 2020-06-15
ability to set chrome process environment variables

bresam:
This gives the ability to set environment variables for the chromium process.


src/Browser/BrowserProcess.php
```diff
@@ -83,8 +83,9 @@ public function __construct(LoggerInterface $logger = null)
      * Starts the browser
      * @param $binary
      * @param $options
+     * @param $envs
      */
-    public function start($binary, $options)
+    public function start($binary, $options, $envs)
```
peter279k, line 88: I think this change should be verified.

GrahamCampbell:
Closing due to inactivity.

PR #179, opened 2020-10-09
Handle target destroyed too fast

momala454:
By example navigating to https://browserleaks.com/canvas
and sending request on the event TargetCreated is crashing the library 

```php
$this->browser->getConnection()->on('method:Target.targetCreated', function (array $params) {
    $this->browser->getTarget($params['targetInfo']['targetId'])->getSession()->sendMessageSync(new HeadlessChromium\Communication\Message('Network.setUserAgentOverride', ['userAgent' => 'abcdef', 'platform' => 'Win32']));
});
```

causes
`Uncaught TypeError: Argument 2 passed to HeadlessChromium\Communication\Session::__construct() must be of the type string, null given, called in /....chrome-php/chrome/src/Communication/Connection.php on line 244`

This pull request detect that the call to attachToTarget failed, because It returns : 
`"error":{"code":-32602,"message":"No target with given id found"}`


GrahamCampbell:
Excellent, thanks.

PR #189, opened 2021-01-27
fixes phpunit assertRegExp deprecation

enricodias:
Deprecation warning from PHPUnit:

> assertRegExp() is deprecated and will be removed in PHPUnit 10. Refactor your code to use assertMatchesRegularExpression() instead.


GrahamCampbell:
This breaks the tests on PHP 7.1. We can just live with the deprecation warning, for now. We have no reason to upgrade to PHPUnit 10 yet. :)

PR #190, opened 2021-01-27
[0.10] Added mouse scrolling

### LINKED ISSUE #83: Scrolling
What is the best method to scroll a page with the library? I have tried addPreScript and addScriptTag to try and inject javascript and then run it, but without success.

enricodias:
Adds a scroll option in the mouse input. This can be used to scroll not only pages, but elements inside pages such as iframes.

Resolves #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -455,11 +455,7 @@ You can emulate the mouse wheel to scroll up and down in a page, frame or elemen
 ```php
 $page->mouse()
     ->scrollDown(100) // scroll down 100px
-    ->scrollUp(50);   // scroll up 50px 
-
-// scrolling doesn't happen instantly like mouse moves
-// the browser may need some time to complete the action
-sleep(1);
+    ->scrollUp(50);   // scroll up 50px
 ```
 
 ### Keyboard API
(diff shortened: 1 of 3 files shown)


enricodias:
This time the tests worked here even in the older versions of php. I'll try to take a look and fix it.

GrahamCampbell:
This seems inherently flakey, so not useful even for someone using this library. Is there a way to block, waiting for chrome to finish the render?

enricodias:
@GrahamCampbell I didn't find a better solution to this, but it worked every time in all my machines even without the `usleep`. I implemented this feature because I needed it for a project. I also didn't find any way to block the execution until the scroll finishes, do you have any suggestion?

GrahamCampbell:
Maybe you need to wait before scrolling?

enricodias:
@GrahamCampbell How do you feel about just printing a warning instead of an error if this test fail? It will prevent breaking the tests on github and will still warn developers about what happened.

enricodias:
Waiting before scrolling seems to work, I should have thought about that before. 😁

GrahamCampbell:
Great, thanks.

PR #203, opened 2021-04-07
Add parameter captureBeyondViewport to screenshot

momala454:
New parameter to allow to capture screenshot beyond the viewport (allowing full page screenshot) without needing to resize it.

```php
$page->screenshot(['captureBeyondViewport' => true, 'clip' => $page->getFullPageClip()])
```


GrahamCampbell:
Thanks. :)

PR #206, opened 2021-04-27
Fix: Auto-detect chrome binary path on MacOS

CViniciusSDias:
On MacOS systems, the chrome binary path is different. This PR aims to fix that issue.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": ">=7.2",
+        "php": "^7.1.3 || ^8.0",
         "apix/log": "^1.2",
         "evenement/evenement": "^3.0.1",
         "psr/log": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^7.1.3 || ^8.0",
+        "php": ">=7.2",
         "apix/log": "^1.2",
         "evenement/evenement": "^3.0.1",
         "psr/log": "^1.0",
--- a/tests/BrowserFactoryTest.php
+++ b/tests/BrowserFactoryTest.php
@@ -26,7 +26,7 @@ class BrowserFactoryTest extends BaseTestCase
 
         $browser = $factory->createBrowser();
 
-        $this->assertRegExp('#^ws://#', $browser->getSocketUri());
+        $this->assertMatchesRegularExpression('#^ws://#', $browser->getSocketUri());
     }
 
     public function testWindowSizeOption()
--- a/tests/BrowserFactoryTest.php
+++ b/tests/BrowserFactoryTest.php
@@ -26,7 +26,7 @@ class BrowserFactoryTest extends BaseTestCase
 
         $browser = $factory->createBrowser();
 
-        $this->assertMatchesRegularExpression('#^ws://#', $browser->getSocketUri());
+        $this->assertRegExp('#^ws://#', $browser->getSocketUri());
     }
 
     public function testWindowSizeOption()
--- a/src/BrowserFactory.php
+++ b/src/BrowserFactory.php
@@ -25,27 +25,20 @@ class BrowserFactory
 
     public function __construct(string $chromeBinary = null)
     {
-        $this->setChromeBinary($chromeBinary);
+        $this->chromeBinary = $chromeBinary ?? $this->getDefaultChromeBinaryPath();
     }
 
-    private function setChromeBinary(?string $chromeBinary): void
+    private function getDefaultChromeBinaryPath(): string
     {
-        if ($chromeBinary !== null) {
-            $this->chromeBinary = $chromeBinary;
-            return;
-        }
-
         if (array_key_exists('CHROME_PATH', $_SERVER)) {
-            $this->chromeBinary = $_SERVER['CHROME_PATH'];
-            return;
+            return $_SERVER['CHROME_PATH'];
         }
 
-        if (PHP_OS_FAMILY === 'Darwin') {
-            $this->chromeBinary = '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
-            return;
+        if (PHP_OS === 'Darwin') {
+            return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
         }
 
-        $this->chromeBinary = 'chrome';
+        return 'chrome';
     }
 
     /**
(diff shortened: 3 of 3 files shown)


composer.json
```diff
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^7.1.3 || ^8.0",
+        "php": ">=7.2",
```
GrahamCampbell: Please revert this change.

tests/BrowserFactoryTest.php
```diff
@@ -26,7 +26,7 @@ public function testBrowserFactory()
 
         $browser = $factory->createBrowser();
 
-        $this->assertRegExp('#^ws://#', $browser->getSocketUri());
+        $this->assertMatchesRegularExpression('#^ws://#', $browser->getSocketUri());
```
GrahamCampbell: Please don;t change this.

GrahamCampbell (changes requested):

tests/BrowserFactoryTest.php
CViniciusSDias: `assertRegExp` is deprecated. Why not to change it?

composer.json
CViniciusSDias: `PHP_OS_FAMILY` constant is available from PHP 7.2 up. I can change the code to use `PHP_OS`. I'll work on that tonight.

GrahamCampbell: Pull in the symfony php 7.2 polyfill, please.

tests/BrowserFactoryTest.php
GrahamCampbell: Because it won;t work on php 7.1.

enricodias:
If we're going to add a feature to auto discover the chrome binary based on the OS, wouldn't it be better to have a separate method to do the discovery instead of the current `setChromeBinary`? This method is doing 2 things.

Maybe something like this would be better:

```php
    public function __construct(string $chromeBinary = null)
    {
        $this->chromeBinary = $chromeBinary ?? $this->getDefaultChromeBinaryPath();
    }

    private function getDefaultChromeBinaryPath(): void
    {
        if (array_key_exists('CHROME_PATH', $_SERVER)) {
            return $_SERVER['CHROME_PATH'];
        }

        if (PHP_OS_FAMILY === 'Darwin') {
           return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
        }

        return 'chrome';
    }
```

The method could be modified to work with Windows later on, or even query the OS to find the correct path.

CViniciusSDias:
> The method could be modified to work with Windows later on, or even query the OS to find the correct path.

AFAIK, `chrome` works fine on _Windows_

tests/BrowserFactoryTest.php
CViniciusSDias: I checked PHPUnit's code and I did not find anything that would make `assertMatchesRegularExpression` fail on PHP 7.1.

I'm having some docker problems so I'm not being able to reproduce the environment. Because of that I'll revert this change, but I still don't understand what would break on PHP 7.1

composer.json
CViniciusSDias: I used `PHP_OS` since we have the same result without an extra dependency.

GrahamCampbell (changes requested):
Pull in the symfony php 7.2 polyfill, please.

CViniciusSDias:
> Pull in the symfony php 7.2 polyfill, please.

What's the reason for that? If it's just for `PHP_OS_FAMILY` I'd rather not since for MacOS the result is exactly the same.

GrahamCampbell:
Oh yes, I didn't spot that change.

enricodias:
> AFAIK, `chrome` works fine on _Windows_

It doesn't, at least on my machines. I have to use the full path.

GrahamCampbell:
What's the most common path to 64-bit chrome on Windows? We might as well throw that in here too.

CViniciusSDias:
> What's the most common path to 64-bit chrome on Windows? We might as well throw that in here too.

Apparently `C:\Program Files (x86)\Google\Chrome\Application\Chrome.exe`. Could you please confirm that, @enricodias ?

enricodias:
@GrahamCampbell it varies a lot. On my current machine it's in `Program Files (x86)\Google\Chrome\Application\chrome.exe` but the "Program Files" folder itself gets translated depending on the installed language. The full path also varies between different Windows versions.

It's probably better to find the correct path using the Windows registry somehow.

CViniciusSDias:
The "Program Files" folder only has alias in the Explorer, but it's always "Program Files" from the CLI. About Windows versions, I don't believe that changes...

enricodias:
@CViniciusSDias On Windows 7 it's in `C:\Program Files (x86)\Google\Application\chrome.exe`, on Vista I believe it was in `C:\Users\UserName\AppDataLocal\Google\Chrome`. I also think that I've seen it in `LocalAppData` before aswell.

Using "Program Files (x86)" worked here in the CLI, but you'll also need to guess the drive letter used. Mine for example wasn't using the default `c:/`

CViniciusSDias:
Ah, I didn't take into consideration the possibility of using something different than `C:`.
Do you know how to get the path from the registry, @enricodias ?
If not, this will probably need some further investigation and I believe it would be better to have it in a different PR, @GrahamCampbell

CViniciusSDias:
Also, I believe we'd need an extension to get values from Windows registry via PHP...

enricodias:
Using the env variable `%ProgramFiles(x86)%` also worked here in the old cmd, but it doesn't return the correct drive letter.

I'll have to take a look on how to query the registry using php without admin permissions. I'm not 100% sure if it's possible.

enricodias:
`start chrome` also worked in the cmd, but it doesn't work as a path to start the process.

I found the [amphp/windows-registry](https://github.com/amphp/windows-registry) package that seems to be able to read keys from the registry. I'll test it later.

GrahamCampbell:
> On Windows 7 it's in C:\Program Files (x86)\Google\Application\chrome.exe

That is not 64-bit chrome?

enricodias:
> That is not 64-bit chrome?

Even though the chrome download page says 64bit, the automated installer always used the x86 version for me. The installer itself doesn't show any options during the setup. That's one more thing to guess; even if the user is using a 64bit system, chrome could be x86.

GrahamCampbell:
Thanks.

PR #208, opened 2021-05-01
Better binary path detection based on user OS

enricodias:
- Adds auto-detect binary path on Windows using the registry.

Following up on #206 

Even tough there are packages built specifically to handle queries to the Windows Registry, I think that adding a new dependence just for that single query is not worth it. I opted to do a simple `shell_exec` and if that fails, guess the path using an env variable.

The new class `AutoDiscover` takes this logic out of the `BrowserFactory` and makes future changes easier.

The registry query was tested only on Windows 10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,10 +83,10 @@ try {
 
 ### Using different chrome executable
 
-When starting the factory will look for the environment variable ``"CHROME_PATH"`` to find the chrome executable.
-If the variable is not found then it will use ``"chrome"`` as the executable.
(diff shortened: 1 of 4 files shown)


enricodias:
The tests worked locally. I'm trying to find out why they are failing here.

enricodias:
The first test was modifying the `$_SERVER['CHROME_PATH']` and making the rest of the test suite fail. I didn't catch this error locally because I was running the tests in parallel, decoupled from each other.

GrahamCampbell:
Thanks. 🍻

PR #212, opened 2021-05-06
Downloads

alain-guidon:
Added a page method to set a path to save downloaded files.

Details found in
https://fahri.id/posts/file-downloading-in-headless-chrome-using-chromedriver-and-hound/


GrahamCampbell:
Thanks. :)

PR #227, opened 2021-06-14
Fix getChromeVersion on windows

### LINKED ISSUE #214: have to manually shut down chromium to run normally
Windows 10 / 64 bit
PHP 7.2
Chromium Version 90.0.4427.0 (Developer Build) (64-bit)

```
require_once __DIR__ . "/vendor/autoload.php";

$browserFactory = new BrowserFactory( __DIR__ ."/browser/chrome-win/chrome.exe" );

/**
 * At this moment, browser will open but never navigate to url or never run anything.
 * Browser will idle , this like some bugs, because at this moment debug logger not starting.
 * i mark this jobs as "jobs A"
**/
	'windowSize'                    => 	[860, 680],
	'headless'		        => 	false,
	'ignoreCertificateErrors'	=>      true,
	'userAgent'		        =>	'Mozilla/5.0 (compatible; MSIE 10.0; Windows Phone 8.0; Trident/6.0; IEMobile/10.0; ARM; Touch; NOKIA; Lumia 920)',
	'debugLogger'                   =>      'php://stdout'
]);

$page->navigate( 'https://example.com' )->waitForNavigation();

$browser->close();
```

as I mentioned, there is a browser that is open but does nothing, and also until now the debuglogger is not running or showing nothing.

but when I close the browser that didn't do anything earlier, without waiting for a long time a new browser will appear and everything works normally.

everything is normal, the debuglogger displays information, the browser starts to open the url I want.

my question is, why do I have to close the first browser so that the second browser appears and runs normally.

I've done everything possible, tried local chrome and others. but everything remains the same, I have to close the first browser so that the second browser appears and runs normally.

is there an error that i don't know about? or does this framework have bugs

enricodias:
The chrome `--version` flag is currently not supported on Windows. This fix uses [wmic](https://docs.microsoft.com/en-us/windows/win32/wmisdk/wmic) as a workaround to get the version directly from the .exe file.

No tests added for this specific case since it would need to run on Windows. A test mocking the `shell_exec` result wouldn't be very useful.

Fixes #214

### DIFF
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -19,15 +19,12 @@ class AutoDiscover
             return $_SERVER['CHROME_PATH'];
         }
 
-        switch ($this->getOS()) {
-            case 'Darwin':
-                return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
-            break;
-            case 'WIN32':
-            case 'WINNT':
-            case 'Windows':
-                return self::windows();
-            break;
+        if ($this->isMac() === true) {
+            return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
+        }
+
+        if ($this->isWindows() === true) {
+            return self::windows();
         }
 
         return 'chrome';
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Dropping this function entirely.

enricodias:
Will the `getChromeVersion` be dropped or just the windows support?

GrahamCampbell:
> getChromeVersion

Just this function is being dropped. The other issue this function caused with incompatibility with the keepAlive option which was confusing some people.

GrahamCampbell:
v0.11.0 has just been released. :)

PR #234, opened 2021-07-06
Prevent symfony process class to kill chrome

Entretoize:
This just extend symfony Process class not to kill chrome at the end of php script execution, it's easy and that works !

### DIFF
--- a/src/BrowserFactory.php
+++ b/src/BrowserFactory.php
@@ -19,6 +19,10 @@ use HeadlessChromium\Exception\BrowserConnectionFailed;
 use Symfony\Component\Process\Process;
 use Wrench\Exception\HandshakeException;
 
+class PersistentProcess extends Process {
+    public function __destruct() {}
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Extending the class is not the right approach here, as I explained in the other issue.

GrahamCampbell:
We should call `start` instead of `run` two lines down.

GrahamCampbell:
We'd also have to share the process instance in that class. Otherwise, we'll still just keep making new ones.

Entretoize:
Then do it with start instead of run, the problem is there from months or maybe years.
I don't understand why it is not the right way, it works for me...

GrahamCampbell:
I am going to remove this version function entirely. If you use `'keepAlive' => true`, then everything should already work.

GrahamCampbell:
v0.11.0 has just been released. :)

PR #239, opened 2021-07-18
Add DOM element locator

### LINKED ISSUE #225: Click on HTML element
Hello world!
I want to developp maybe a new feature for this wonderful project.
I was thinking that maybe 70% of the time we use this library form fill and submit form. 
I was thinking at a functionality to directly click on a html element in php.
Maybe something like `$page->click('.my-selector');`!
We can add this functionnality with a lot of function like :
- form submit `$page->submitForm('.my-selector');`
- fill input `$page->fillInput('.my-selector', 'some text');`*
- etc.

And maybe with a new object like that `$page->formHandler->fillInput('.my-selector');`

And maybe after do a whole html support with node, etc.

Thanks for reading this!
(and sorry for my bad english)

enricodias:
Allows searching elements using the new `find` method. Further work is needed to make it work on edge cases, such as if the element is outside the visible area to the left or top.

Resolves #225

### DIFF
--- a/src/Input/Mouse.php
+++ b/src/Input/Mouse.php
@@ -12,6 +12,8 @@
 namespace HeadlessChromium\Input;
 
 use HeadlessChromium\Communication\Message;
+use HeadlessChromium\Exception\ElementNotFoundException;
+use HeadlessChromium\Exception\JavascriptException;
 use HeadlessChromium\Page;
 use HeadlessChromium\Utils;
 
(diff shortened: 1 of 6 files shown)


enricodias:
It's incredible how the StyleCI finishes running before my browser manages to load the page :)

enricodias:
I'll have to rework those scroll methods in another PR. Having to increase the sleep time in the tests isn't a good solution.

enricodias:
Finally, all tests passed.

Sorry for the mess with all those commits. It seems like these new tests with Symfony run slower than before. I couldn't reproduce that behavior locally and had to keep pushing here.

GrahamCampbell:
Sounds good. Let's merge that other PR into 0.11, then I can merge upwards into 1.0, then this PR can be rebased (and kept targetting 1.0).

AlexeyKosov:
Is there anything blocking this PR from merging? Looking forward to using the new functionality

GrahamCampbell:
Are you able to rebase this @enricodias. I am going to do the 1.0.0 release now, so this PR will be for 1.1.0. :)

enricodias:
@GrahamCampbell Done. I also removed the `usleep` calls since they are no longer needed.

GrahamCampbell:
🎉

src/Input/Mouse.php
```diff
@@ -207,6 +209,95 @@ private function scroll(int $distanceY, int $distanceX = 0): self
...
+
+        try {
+            $elementList = $this->page
+                ->evaluate('JSON.parse(JSON.stringify(document.querySelectorAll("'.$selectors.'")));')
+                ->getReturnValue();
+
+            $position = \max(0, (\count($elementList) - 1));
+
+            $element = $this->page
+                ->evaluate('JSON.parse(JSON.stringify(document.querySelectorAll("'.$selectors.'")['.$position.'].getBoundingClientRect()));')
```
momala454, line 276: shouldn't we escape the selector to escape " ?

enricodias: You mean to accept things like `->find('input[name="name"]');`?

momala454: yes

PR #240, opened 2021-07-18
Fix scroll method returning prematurely 

enricodias:
The `scroll` method will return only when the browser finishes scrolling. Removing the need of using `\usleep` in tests.

### DIFF
--- a/src/Input/Mouse.php
+++ b/src/Input/Mouse.php
@@ -13,6 +13,7 @@ namespace HeadlessChromium\Input;
 
 use HeadlessChromium\Communication\Message;
 use HeadlessChromium\Page;
+use HeadlessChromium\Utils;
 
 class Mouse
 {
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Please can you send this to the 0.11 branch?

src/Input/Mouse.php
```diff
@@ -181,13 +193,64 @@ private function scroll(int $distance)
...
+    private function waitForScroll(int $targetX, int $targetY)
+    {
+        while (true) {
+            $visibleArea = $this->page->getLayoutMetrics()->getVisualViewport();
+
+            if ($visibleArea['pageX'] === $targetX && $visibleArea['pageY'] === $targetY) {
+                return true;
+            }
+
+            yield 1000;
```
GrahamCampbell, line 257: what does this do?

GrahamCampbell: (the `1000`, not the `yield` :trollface:)

GrahamCampbell (changes requested):

src/Input/Mouse.php
enricodias: It's the number of microseconds to wait before retry, for the `\HeadlessChromium\Utils::tryWithTimeout` method. I'll update the docblock.

enricodias:
> Please can you send this to the 0.11 branch?

Done.

PR #247, opened 2021-08-15
Change `bellow` to `below`

buismaarten:
According to https://prowritingaid.com/grammar/1000167/Bellow-vs-below%E2%80%94what-is-the-difference, `bellow` has a different meaning than `below`.


GrahamCampbell:
Thanks. :)

PR #259, opened 2021-09-23
Added interacting with dom

Seriyyy95:
This pull request allows interact with page DOM, find elements by css selector or xpath, scroll to elements, click on it, get elements text and attributes, upload files.

### DIFF
--- a/src/Dom/Dom.php
+++ b/src/Dom/Dom.php
@@ -9,9 +9,7 @@ use HeadlessChromium\Page;
 
 class Dom extends Node
 {
-    /**
-     * Dom constructor.
-     */
+
     public function __construct(Page $page)
     {
         $message = new Message('DOM.getDocument');
--- a/src/Dom/Dom.php
+++ b/src/Dom/Dom.php
@@ -9,7 +9,6 @@ use HeadlessChromium\Page;
 
 class Dom extends Node
 {
-
     public function __construct(Page $page)
     {
         $message = new Message('DOM.getDocument');
--- a/src/Dom/Dom.php
+++ b/src/Dom/Dom.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace HeadlessChromium\Dom;
+
+use HeadlessChromium\Communication\Message;
+use HeadlessChromium\Page;
+
+/**
+ * Class Dom
+ */
+class Dom extends Node
+{
+
+    /**
+     * @param Page $page
+     */
+    public function __construct($page)
+    {
+        $message = new Message('DOM.getDocument');
+        $stream = $page->getSession()->sendMessage($message);
+        $response = $stream->waitForResponse(1000);
+
+        $rootNodeId = $response->getResultData('root')['nodeId'];
+
+        parent::__construct($page, $rootNodeId);
+    }
+    
+    public function search($selector)
+    {
+        $message = new Message('DOM.performSearch', [
+            'query' => $selector
+        ]);
+        $response = $this->page->getSession()->sendMessageSync($message);
+
+        $this->assertNotError($response);
+        
+        $searchId = $response->getResultData('searchId');
+        $count =  $response->getResultData('resultCount');
+
+        if($count === 0){
+            return [];
+        }
+        $message = new Message('DOM.getSearchResults', [
+            'searchId' => $searchId,
+            'fromIndex' => 0,
+            'toIndex' => $count
+        ]);
+
+        $response = $this->page->getSession()->sendMessageSync($message);
+
+        $this->assertNotError($response);
+        
+        $nodes = [];
+        $nodeIds = $response->getResultData('nodeIds');
+        foreach ($nodeIds as $nodeId) {
+            $nodes[] = new Node($this->page, $nodeId);
+        }
+
+        return $nodes;
+    }
+
+}
--- a/src/Dom/Dom.php
+++ b/src/Dom/Dom.php
@@ -7,16 +7,13 @@ namespace HeadlessChromium\Dom;
 use HeadlessChromium\Communication\Message;
 use HeadlessChromium\Page;
 
-/**
- * Class Dom.
- */
 class Dom extends Node
 {
 
     /**
-     * @param Page $page
+     * Dom constructor.
      */
-    public function __construct($page)
+    public function __construct(Page $page)
     {
         $message = new Message('DOM.getDocument');
         $stream = $page->getSession()->sendMessage($message);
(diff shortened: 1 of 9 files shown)


enricodias:
Interesting.

Tests and docs are missing. Lot's of methods doesn't have defined types for the args and return value.

Seriyyy95:
> Tests and docs are missing. Lot's of methods doesn't have defined types for the args and return value.

Ok. I will add docs, tests and return&arguments types.

src/Page.php
```diff
@@ -989,9 +993,9 @@ public function setCookies($cookies)
      *
      * @param string $userAgent
      *
+     * @return ResponseWaiter
```
GrahamCampbell: Please don't change the order here.

GrahamCampbell (changes requested):

src/Dom/Node.php
```diff
@@ -0,0 +1,237 @@
...
+
+        $attributes = $response->getResultData('attributes');
+
+        return new NodeAttributes($attributes);
+    }
+
+    /**
+     * @param string $selector
+     *
+     * @return Node|void
```
GrahamCampbell: Please don't use void in a union. return null instead

src/Dom/Node.php
```diff
@@ -0,0 +1,237 @@
...
+declare(strict_types=1);
+
+namespace HeadlessChromium\Dom;
+
+use HeadlessChromium\Communication\Message;
+use HeadlessChromium\Exception\DomException;
+use HeadlessChromium\Page;
+
+/**
+ * Class Node
```
GrahamCampbell: useless comment. remove please.

src/Dom/Dom.php
```diff
@@ -0,0 +1,63 @@
...
+
+declare(strict_types=1);
+
+namespace HeadlessChromium\Dom;
+
+use HeadlessChromium\Communication\Message;
+use HeadlessChromium\Page;
+
+/**
+ * Class Dom.
```
GrahamCampbell: same

src/Dom/Dom.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+class Dom extends Node
+{
+
+    /**
+     * @param Page $page
```
GrahamCampbell: please just add the real type here to the arg

src/Dom/Dom.php
```diff
@@ -0,0 +1,63 @@
...
+        $message = new Message('DOM.getDocument');
+        $stream = $page->getSession()->sendMessage($message);
+        $response = $stream->waitForResponse(1000);
+
+        $rootNodeId = $response->getResultData('root')['nodeId'];
+
+        parent::__construct($page, $rootNodeId);
+    }
+
+    public function search($selector)
```
GrahamCampbell: missing type

GrahamCampbell (changes requested):

Seriyyy95:
I fixed all of this. Should I fix anything else?

src/Dom/Dom.php
GrahamCampbell: please add a real type here, instead of just phpdoc

GrahamCampbell (changes requested):

Seriyyy95:
I added real types for return. Should I add real types for args? There is no real arg types in other code (except contructor).

src/Dom/Dom.php
```diff
@@ -29,7 +28,7 @@ public function __construct(Page $page)
      *
      * @return array
      */
-    public function search($selector)
+    public function search($selector): array
```
GrahamCampbell: `string $selector`

GrahamCampbell (changes requested):

GrahamCampbell:
Yeh. Wanna make new code strongly typed. :)

Seriyyy95:
Done.

enricodias:
An unrelated test is failing with this PR, and it's one of my tests :/

>1) HeadlessChromium\Test\MouseApiTest::testFind_withScrolling
HeadlessChromium\Exception\OperationTimedOut: Operation timed out (30sec)

I'll take a look as soon as I can.

src/Dom/Dom.php
```diff
@@ -0,0 +1,64 @@
...
+
+namespace HeadlessChromium\Dom;
+
+use HeadlessChromium\Communication\Message;
+use HeadlessChromium\Page;
+
+class Dom extends Node
+{
+    /**
+     * Dom constructor.
```
GrahamCampbell: Please don't include comments like this that don't add any extra information. I know things like this may be present on existing code, but I'd like to make sure new code is of a better quality. :)

src/Dom/Dom.php
```diff
@@ -0,0 +1,64 @@
...
+        $stream = $page->getSession()->sendMessage($message);
+        $response = $stream->waitForResponse(1000);
+
+        $rootNodeId = $response->getResultData('root')['nodeId'];
+
+        parent::__construct($page, $rootNodeId);
+    }
+
+    /**
+     * @param string $selector
```
GrahamCampbell: please remove any phpdoc that duplicates exactly the real types.

GrahamCampbell (changes requested):

enricodias:
Sorry for the delay.

I couldn't reproduce the test error locally running php8 with symfony4. Maybe the chrome process died unexpectedly during the workflow. Let's see if it fails again in the next push.

src/Dom/Dom.php
Seriyyy95: PHPDoc comments with duplicates removed.

src/Page.php
```diff
@@ -776,7 +776,7 @@ public function keyboard()
     }
 
     /**
-     * @return Dom
+     * Get Dom object
```
GrahamCampbell: just delete this doc

GrahamCampbell (changes requested):

src/Page.php
Seriyyy95: These comments already deleted.

Seriyyy95: Do you want me to delete "Get Dom object" too?

GrahamCampbell: yes, doesn't add any value. the function is called dom and typed with `Dom` :trollface:

Seriyyy95: Done

enricodias:
Well, at least it didn't fail in the php8 test again 😅

Seriyyy95:
Tests fails becasue PHP 7.3 does not support typed class properties.

GrahamCampbell:
Please replace property typing with phpdoc.

Seriyyy95:
Done.

enricodias (approved):
lgtm if the tests pass.

Seriyyy95:
Tests fail not on my code. There is problem with method testConnectToBrowser, I can reproduce it on local machine, but dont know how to fix it.

GrahamCampbell:
Thanks! 🍻

PR #278, opened 2021-10-27
Add `HeadlessChromium\Page::setTimezone` method

octoper:
This PR adds `HeadlessChromium\Page::setTimezone` method ([Emulation.setTimezoneOverride](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setTimezoneOverride))

### DIFF
--- a/src/Page.php
+++ b/src/Page.php
@@ -18,6 +18,7 @@ use HeadlessChromium\Cookies\Cookie;
 use HeadlessChromium\Cookies\CookiesCollection;
 use HeadlessChromium\Dom\Dom;
 use HeadlessChromium\Exception\CommunicationException;
+use HeadlessChromium\Exception\InvalidTimezoneId;
 use HeadlessChromium\Exception\NoResponseAvailable;
 use HeadlessChromium\Exception\TargetDestroyed;
 use HeadlessChromium\Input\Keyboard;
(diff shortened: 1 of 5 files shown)


tests/PageTest.php
```diff
@@ -57,6 +58,72 @@ public function testSetUserAgent(): void
...
+            $page->setTimezone('Foo/Bar');
+        } catch (\Exception $e) {
+            $this->assertEquals('Invalid Timezone ID: Foo/Bar', $e->getMessage());
+        }
+
+        try {
+            $page->setTimezone('Baz/Qux');
+        } catch (\Exception $e) {
+            $this->assertEquals('Invalid Timezone ID: Baz/Qux', $e->getMessage());
+        }
```
enricodias: You can use `$this->expectException(InvalidTimezoneId::class);` to tell phpunit what exception you are expecting, then you don't have to use try{} catch{}

octoper: Done!

octoper:
Hey @enricodias this is done if you want to check it again! (sorry for the ping)

enricodias:
I saw it, but the styleci check has some errors. Can you fix it to make the pr mergeble?

src/Page.php
```diff
@@ -811,6 +812,33 @@ public function assertNotClosed(): void
...
+            ->sendMessageSync(
+                new Message(
+                    'Emulation.setTimezoneOverride',
+                    [
+                        'timezoneId' => $timezoneId ?? '',
+                    ]
+                )
+            );
+
+        if (mb_strpos($response->getErrorMessage(), 'Invalid timezone')) {
```
GrahamCampbell: why mb_strpos?

GrahamCampbell (changes requested):

src/Page.php
octoper: Because the error message has some more characters as prefix

GrahamCampbell: i mean as opposed to strpos

GrahamCampbell: we don't depend on the mbstring extension

octoper: Oh I think there is no difference so I can update to use `strpos`

GrahamCampbell (approved):

enricodias (approved):

GrahamCampbell:
Thanks. :)

PR #283, opened 2021-11-07
Rework PagePdf class to improve validations and tests

### LINKED ISSUE #275: Print to PDF - specify pages?
Is there a way to specify which pdf pages should be created? E.g. if I only want to print first page (so that pdf file would have first page only), how can I do that?

Chrome supports 'pageRanges' option, but not sure if that's implemented here? Tried that, but it didn't work

https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF

enricodias:
- The validations were improved and moved to the `PagePdf::class`, making it easier to add new options to the list.
- `pageRanges` and `ignoreInvalidPageRanges` were added to the options list (in a separate commit).
- Tests were introduced to test both the validations and the pdf creation.

I'm using `is_numeric` to validate numbers that were previously validated as float/double. The doc says they must be a number, not necessarily a float.

Resolves #275

### DIFF
--- a/src/PageUtils/PagePdf.php
+++ b/src/PageUtils/PagePdf.php
@@ -36,7 +36,7 @@ class PagePdf extends AbstractBinaryInput
         'marginBottom' => self::TYPE_NUMERIC,
         'marginLeft' => self::TYPE_NUMERIC,
         'marginRight' => self::TYPE_NUMERIC,
-        'pageRanges'              => self::TYPE_STRING,
+        'pageRanges' => self::TYPE_STRING,
         'ignoreInvalidPageRanges' => self::TYPE_BOOLEAN,
         'preferCSSPageSize' => self::TYPE_BOOLEAN,
         'scale' => self::TYPE_NUMERIC,
--- a/src/PageUtils/PagePdf.php
+++ b/src/PageUtils/PagePdf.php
@@ -11,10 +11,104 @@
 
 namespace HeadlessChromium\PageUtils;
 
+use HeadlessChromium\Communication\Message;
 use HeadlessChromium\Exception\PdfFailed;
+use HeadlessChromium\Page;
 
 class PagePdf extends AbstractBinaryInput
 {
+    private const TYPE_NUMERIC = 1;
+    private const TYPE_STRING = 2;
+    private const TYPE_BOOLEAN = 3;
+
+    /**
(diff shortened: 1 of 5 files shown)


enricodias:
Interesting how StyleCI doesn't like the alignment on arrays but requires it in the doc blocks.

GrahamCampbell:
That code style choice is not a StyleCI design choice, it's a choice made by us. https://github.com/chrome-php/chrome/blob/1.1/.styleci.yml

src/PageUtils/PagePdf.php
```diff
@@ -26,4 +120,12 @@ protected function getException(string $message): \Exception
             \sprintf('Cannot make a PDF. Reason : %s', $message)
         );
     }
+
+    /**
+     * Wrapper to throw exception in expression when running in php 7.
+     */
```
GrahamCampbell, line 128: missing throws doc

src/PageUtils/PagePdf.php
```diff
@@ -11,10 +11,104 @@
...
+    {
+        \array_map([$this, 'validateOption'], \array_keys($options), $options);
+
+        $this->options = $options;
+
+        return $this;
+    }
+
+    /**
+     * @param string                $name  Option name
```
GrahamCampbell: don't think the option name and value comments add any value. the fuction is called validateOption and has name and value params, so :trollface:

GrahamCampbell (approved):

src/PageUtils/PagePdf.php
enricodias: It's to specify the types of `$value` since union types are not supported in php 7. It felt wrong to specify only one of the params. I can remove the descriptions and leave only the types if you think it's better.

GrahamCampbell: i was referring specifically to the text `Option name`, not the whole line. ;)

GrahamCampbell (changes requested):
clicked the wrong button last time :trollface:

src/PageUtils/PagePdf.php
enricodias: Ok, I'll amend the commit later today when I get home.

enricodias:
All done.

GrahamCampbell (approved):

PR #293, opened 2021-11-23
Added support for setting HTTP headers

### LINKED ISSUE #253: Add built in support for setting additional headers
The one thing this project is lacking, is an easy way to set/pass additional headers.

I think that on the browser factory creation, we should be able to pass a headers array. Something like this:

```php
    'userAgent' => $userAgent,
    'windowSize' => [1920, 1000],
    'headers' => [
        'Accept-Language' => 'en-US,en;q=0.9',
    ],
]);
```

It would also help to have a helper function as well, like:

```php
$browserFactory->addHeader('key', 'value');
```

and/or

```php
$browserFactory->addHeaders($array);
```

I managed to get the headers to work by passing them on the page instance. It just seems like it would make more sense to pass them on the initial browser instance as a global setting.

Here is how I added them to the page instance:

```php

$page->getSession()->sendMessageSync(new Message('Network.setExtraHTTPHeaders', [
    'headers' => [
        'Accept-Encoding' => 'identity',
        'Accept-Language' => 'en-US,en;q=0.9',
    ]
]));

$page->navigate($url);
```

maximgrynykha:
Resolve #253


maximgrynykha:
@enricodias Can you look at changes and help with tests?

enricodias:
I'll take a look at it as soon as I can :)

GrahamCampbell:
As per the contribution guidelines, we follow semantic versioning, so I've re-targetted this to the 1.3 branch.

GrahamCampbell:
>  feat: add the ability to emulate specific devices 

Can you submit this as a separate PR please?

maximgrynykha:
> > feat: add the ability to emulate specific devices
> 
> Can you submit this as a separate PR please?

Yes of course. I was wrong.

maximgrynykha:
I am not sure but seems CI-checks fail not because of my fault.

enricodias:
There's an intermittent test that may result in a timeout, usually in the Symfony 4 test with php 8.0. I don't know the cause yet. If you push again to trigger a new build, it should pass.

I don't know if we should test that a custom header is sent in the requests. We can test that the `addHeader()` method doesn't break anything when called, but ~~we would need a server to test the actual feature. We could use the php built-in serve, but it seems overkill and~~ we would be testing chromium and not our package.

Edit: we could listen to the `Network.requestWillBeSent` event and check the request data, but I didn't manage to make it work yet.

enricodias:
I created a few tests but I don't have write access to commit them directly into this pr 😬

@GrahamCampbell can you give me write access to this branch?

GrahamCampbell:
@enricodias I've changed your role to owner on the organization.

enricodias:
Thanks, Graham.

I added some tests, updated a doc link and modified the method `Page::setBasicAuthHeader` to call the new `setExtraHTTPHeaders` method.

I was also messing with the `$option` cache to merge with the current option list instead of overwriting it, but now I'm having second thoughts. If we just overwrite the options, the addHeader methods won't be useful in cases where we need to set any other option, but merging the way I did will prevent any option value from changing in subsequent calls. What do you guys think? Merge without doing the diff?

ismaxim:
@enricodias 

Thank you for the tests. If you have any additional remarks or ideas about the feature of this pull-request, please discuss.

maximgrynykha:
I support your solution to merging with diffs. In cases, if consumers of the package need to set up "additional" headers for the page, they will do this for the concrete instance of the Page::class by using Page::setExtraHTTPHeaders(). After all, you did the right thing - prevent an overwriting which I haven't seen.

Consumers also can write smth, like this:

```php
    "headers" => [
        "User-Agent" => "some user-agent-1"
    ]
    "headers" => [
        "User-Agent" => "some user-agent-2"
    ]
]);
```
In fact, they will operate only the first headers-occurrence in the array of the options. But it is **important to note this behavior at PHPDocs and README.**

src/BrowserFactory.php
```diff
@@ -63,12 +65,38 @@ public function createBrowser(array $options = []): ProcessAwareBrowser
             \register_shutdown_function([$browserProcess, 'kill']);
         }
 
+        // Cache options at factory
+        $this->options = \array_merge($this->options, $options);
```
maximgrynykha: There is a new change again?

enricodias:
But since the method is public, people will still be able to call it several times in the same object. They won't immediately know that the options used in the first creation persists in all subsequent creations. Take a look at the modifications in the test case in my last commit.

src/BrowserFactory.php
enricodias: I just added a comment about it :)

maximgrynykha: Yep😀

maximgrynykha:
There are two ways:

1. Note [this behavior](https://github.com/chrome-php/chrome/pull/293#issuecomment-981165525) at PHPDocs and README (+ rollback commit changes).
2. Leave it as it is now.

Required to make a choice.

enricodias:
Since this would break backwards compatibility, documenting this behavior won't be enough. It would require a new major version. It's easier to keep it compatible with the previous implementation.

maximgrynykha:
As I mentioned (then have deleted), again (after verifying with test not only by the memory) - merge can be shorter:

```php
$arr1 = [
    'first' => 1,
    'second' => 2
];

$arr2 = [
    'third' => 3
];


$arr1 += $arr2;

print_r($arr1);

// But if the key at the $arr2 will be the same ('second') as at the first ('second') then only the first array key will be used
```

It is more likely the same as you did previously with - ```\array_merge($this->options, \array_diff_key($options,  $this->options));```

maximgrynykha:
Although, that is just for the mention about the ability not the note to implementation. 
And here is Brent's article about the difference - https://stitcher.io/blog/what-is-array-plus-in-php (+1 huge backlink👋)

```...What's happening here is that array_merge will override existing keys, while + will not. In other words: when a key exists in the first array, + will not merge an item with the same key from another array into the first one.```

At the end of the ends, ```\array_merge($this->options, \array_diff_key($options, $this->options))``` can be written the next way ```$this->options += $options```

**Comment by enricodias** [comment]
Yeah, I was just about to comment that. `+=` will not overwrite values while `array_merge` will. The previous implementation using diff was just the first one that came to my mind when I was writing.

I guess we will leave this pr this way then.

**Review by enricodias — APPROVED** [review]

**Comment by maximgrynykha** [comment]
@enricodias

There is still the same point as you had mentioned before with overwriting options.

For example:
```php
$arr1 = [
    'first' => 'a',   
    'second' => 'b'    
];

$arr2 = [
    'second' => 'c',
    'firth' => 'a'
];

print_r(array_merge($arr1, $arr2));
```
[See the code ->](https://github.com/chrome-php/chrome/pull/293#issuecomment-981170159)

maximgrynykha:
> Since this would break backwards compatibility, documenting this behavior won't be enough. It would require a new major version. It's easier to keep it compatible with the previous implementation.

If any consumer of the package already has overwritten one of the options, this PR will possibly break backward compatibility.

enricodias:
Hmm, you mean if someone is doing `createBrowser(['option1' => 1])` and then `createBrowser()` again and expecting no options? Indeed, it will have a different behavior. I guess we could just change to `$options = \array_merge($this->options, $options);`

maximgrynykha:
> Hmm, you mean if someone is doing `createBrowser(['option1' => 1])` and then `createBrowser()` again expected no options? Indeed, it will have a different behavior. I guess we could just change to `$options = \array_merge($this->options, $options);`

🎯

maximgrynykha:
> Hmm, you mean if someone is doing `createBrowser(['option1' => 1])` and then `createBrowser()` again expected no options? Indeed, it will have a different behavior. I guess we could just change to `$options = \array_merge($this->options, $options);`

The next instance which will be created with maybe other option values by the same key may break previous instances with their options. ```$options = \array_merge($this->options, $options)``` is the one way I see to fix that issue.

enricodias:
Changing it to a local variable will make the tests fail since they can't check that the merge was done correctly. I'll sleep on it and think of a more elegant solution tomorrow.

I guess I bothered Graham for nothing by requesting a review from him before the PR was ready. 😅

maximgrynykha:
> Changing it to a local variable will make the tests fail since they can't check that the merge was done correctly. I'll sleep on it and think of a more elegant solution tomorrow.
> 
> I guess I bothered Graham for nothing by requesting a review from him before the PR was ready. 😅

Gathering the gang together 😎

enricodias:
Not the most beautiful solution, but I think it won't break anything now.

Issue #274, opened 2021-10-23
Is it possible to enable devtools and toggle device?


holden321:
I believe it would be better than just set userAgent to emulate some device.

maximgrynykha:
Hi, @holden321 I am already thought about that feature. I see as I could to spend some time to implement this in future releases. Stay in touch for updates.

PR #294, opened 2021-11-24
feat: emulate devices

maximgrynykha:
#274


maximgrynykha:
**Requires tests and detailed review.**

Need to add exceptions if [DeviceFactory::class](https://github.com/ismaxim/chrome/blob/1.3/src/PageUtils/Devices/DeviceFactory.php) does not have the requested device. - DONE

maximgrynykha:
Does anyone think that [Device::class](https://github.com/ismaxim/chrome/blob/1.3/src/PageUtils/Devices/Device.php), ([Resolution::class](https://github.com/ismaxim/chrome/blob/1.3/src/PageUtils/Devices/Resolution.php), [Viewport::class](https://github.com/ismaxim/chrome/blob/1.3/src/PageUtils/Devices/Viewport.php)) needs setters?

maximgrynykha:
I just have looked at the array of the devices at [DeviceFactory::class](https://github.com/ismaxim/chrome/blob/1.3/src/PageUtils/Devices/DeviceFactory.php) and am not sure if each device in the array needs to duplicate these fields:

'isMobile' => true,
'hasTouch' => true,
'isLandscape' => false,

because they are always similar. May we set that fields at the Device by default?

maximgrynykha:
@GrahamCampbell there is a way to me automatically resolve issues from StyleCI. I want to shoot to myself because of any trailing point in the sentence. I mean it needs to be a tool which automatically helps to fix our wrong)

src/Page.php
```diff
@@ -69,6 +70,60 @@ public function __construct(Target $target, array $frameTree)
...
+     * $browserFactory = new BrowserFactory();
+     * 
+     * // starts headless chrome
+     * $browser = $browserFactory->createBrowser();
+     * 
+     * // get the device for emulation
+     * $iphone = $browserFactory->getDevice('Iphone 6');
+     * 
+     * // specify device positioning 
+     * $iphone->landscape();
```
maximgrynykha, line 95: Perhaps it could be useful to return $this from [Device::class](https://github.com/ismaxim/chrome/blob/1.3/src/PageUtils/Devices/Device.php) instance to achieve that syntax:

```php
- $page->emulate($iphone->landscape());
- $page->emulate($iphone->portrait());
```

Instead of this:

```php
$iphone->landscape(); // or $iphone->portrait()
$page->emulate($iphone);
```

What do you think?

src/PageUtils/Devices/DeviceFactory.php
```diff
@@ -0,0 +1,585 @@
...
+    public function __construct()
+    {
+        foreach ($this->devices() as $device) {
+            $deviceViewport = $device['viewport'];
+
+            $resolution = new class($deviceViewport['width'], $deviceViewport['height'], $deviceViewport['deviceScaleFactor']) extends Resolution {};
+        
+            $viewport = new class($resolution, $deviceViewport['isMobile'], $deviceViewport['hasTouch'], $deviceViewport['isLandscape']) extends Viewport {};
+        
+            $device = new class($device['name'], $device['user-agent'], $viewport) extends Device {};
```
maximgrynykha, line 29: Need to be fixed because of variable `$device` collision

src/PageUtils/Devices/DeviceFactory.php
```diff
@@ -0,0 +1,585 @@
...
+            throw new InexistentDevice("Device {$name} isn't supported.");
+        }
+
+        return $this->devices[$name];
+    }
+
+    /**
+     * @return array<string, string|array<string, int|bool>>
+     */
+    private function devices(): array
```
maximgrynykha, line 52: I think it would be better if I remove this array to a separate PHP file - *DeviceDescriptors.php for example, and then include this one inside the DeviceFactory::device() method. What do you think?

src/PageUtils/Devices/DeviceFactory.php
enricodias: I don't see a reason for those 3 different classes. They seem to be just data classes with fairly few values. Why not have just a single class?

enricodias:
I don't think it's a good idea to list devices like this, especially inside a single method. There are thousands of different devices out there with a combination of different user agents and viewports. We can't list them all without bloating the code, not even the most popular ones.

I searched for a dedicated package that could create user agents on the fly but I couldn't find anything good. Maybe we should start by just creating an interface to set a custom viewport and user agent.

maximgrynykha:
> I don't think it's a good idea to list devices like this, especially inside a single method. There are thousands of different devices out there with a combination of different user agents and viewports. We can't list them all without bloating the code, not even the most popular ones.
> 
> I searched for a dedicated package that could create user agents on the fly but I couldn't find anything good. Maybe we should start by just creating an interface to set a custom viewport and user agent.

Living in the perfect world I would prefer to do the next:

 ⁃ scrape/parse resources to collect all available user-agents

 ⁃ scrape/parse resources to collect all available info for each device, such as viewport and maybe a little more data. 

While finished scraping/parsing the data I would sanitize and normalize the data sets with some database for example - SQLite (because of relationships on data). I know that it locks I/O for some time, but SQLite is one of the fastest DBMS for selecting rows. Even if we will try to use files we will have caught the same issue.

After all, some API* endpoints come into play. This could be string REST-like queries or some SDK-like methods to access the data from SQLite. 

__Theoretichally, I believe that it could be some web interface on JS which operates for adding new devices. That web interface needs to host on GitHub Pages deploying from the repository which holds SQLite database or simple JSON. Each new operation with adding a new device call GitHub API endpoint to manage repository - creates a new issue/pull-request for inserting the data.__ If you are proactive enough you may add some SEO and try to propagate this as a separate project) Final notes - for auth you can use Firebase.

But most importantly at all, we need to get one concrete data set to instantiate the device by abstraction (abstract class preferable because some of the default implementation of the “getter”-methods and proper properties for example. At PHP we cannot achieve ditto the interfaces syntax that is in TypeScript (see the next comment at thread)).

maximgrynykha:
> I don't think it's a good idea to list devices like this, especially inside a single method. There are thousands of different devices out there with a combination of different user agents and viewports. We can't list them all without bloating the code, not even the most popular ones.
> 
> I searched for a dedicated package that could create user agents on the fly but I couldn't find anything good. Maybe we should start by just creating an interface to set a custom viewport and user agent.

I understand that we are not in use with TypeScript but if we look at implementation which I had been inspired, the [Puppeteer](https://github.com/puppeteer/puppeteer) - https://github.com/puppeteer/puppeteer/blob/main/src/common/DeviceDescriptors.ts

maximgrynykha:
We can try to write a separate package that will automatically generate User-Agents. But there are a few pitfalls:

- how to be sure if the User-Agent OS will not differentiate with device OS (the devices we can determine by the viewport sizes)?
- how to be sure if the version of OS at User-Agent will not differentiate with OS versions supported by the device?
- where to store the data for combining the parts of User-Agent for the concrete one string? Maybe at files...

src/PageUtils/Devices/DeviceFactory.php
maximgrynykha: > 

I can agree with something at the point of nonsense with these classes in the general purpose of this package. But with the perspective of reusing the abstraction, I had to prefer these classes. I will look at this point later again.

maximgrynykha: And yes, there is a transitional dependency with Device::class and Resolution::class through Viewport::class ( Device::class -> Viewport::class -> Resolution::class) which is not good for Low Coupling. And breaks the [Low of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter)

enricodias:
I didn't know that Puppeteer still maintain a device list hard coded like this. I still think this isn't the best solution and that a dedicated package to creating user agents would be better. This will also allow other projects to use this feature without importing the entire `chrome-php/chrome` package.

A dedicated interface to transfer this data around would also be nice, like [php-extended/php-user-agent-interface.](https://packagist.org/packages/php-extended/php-user-agent-interface). The same creator also created the package [php-extended/php-user-agent-catalog](https://packagist.org/packages/php-extended/php-user-agent-catalog) to create user agents, but it was abandoned.

GrahamCampbell:
Can you rebase this please?

maximgrynykha:
> Can you rebase this, please?

@GrahamCampbell 
I am not sure what do you mean by rebasing. Can you describe what exactly do you mean (is it Git rebase or smth. else)?

maximgrynykha:
I think this pull request may be closed for some time and reopened in the future with a more interesting implementation of the feature.

GrahamCampbell:
Ok, I'll close this for now, if there's no capacity for it to be worked.

PR #296, opened 2021-11-29
Suppress the depreciation error in PHP 8.1

luigel:
To suppress the error for the following files I added ReturnTypeWillChange attributes. The same for the Symfony symfony/symfony#41495

This PR will have a support for PHP8.1

### DIFF
--- a/src/Communication/Response.php
+++ b/src/Communication/Response.php
@@ -101,6 +101,7 @@ class Response implements \ArrayAccess
     /**
      * {@inheritdoc}
      */
+    #[\ReturnTypeWillChange]
     public function offsetExists($offset)
     {
         return \array_key_exists($offset, $this->data);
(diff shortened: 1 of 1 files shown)


enricodias:
@GrahamCampbell should we add new workflows to run the tests with php 8.1 already?

GrahamCampbell:
There's little point at the moment because https://github.com/chrome-php/wrench doesn't work on PHP 8.1 (the unit tests fail). If someone can get to the bottom of that first, it'd be much appreciated. Until then, I'm not keen on merging PHP 8.1 compatibility fixes here yet. It won't work properly until wrench is fixed.

enricodias:
I'll try to take a look at it in the next days.

GrahamCampbell:
FYI I did some investigation a few weeks back, if it helps: https://github.com/chrome-php/wrench/commits/php81-support.

GrahamCampbell:
Gonna merge this anyway, even though we don't support PHP 8.1 yet.

PR #299, opened 2021-12-07
Get an Array of Cookies

carsteneu:
We added 2 methods to get an Array of used Cookies of a given page and an Array of the names of these Cookies, for this purpose we added a method to get the whole raw data from the Cookie Class.

`$cookies->getListNamesAsArray();`
`$cookies->getListAllDataAsArray();`

This is quite handy if you want to have a Look at the Cookies a page ist using.

Also added .idea for the gitignore - setting files from jetbrains IDEs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
+.idea
--- a/src/Cookies/Cookie.php
+++ b/src/Cookies/Cookie.php
@@ -16,7 +16,7 @@ class Cookie implements \ArrayAccess
     /**
      * @var array
      */
-    protected $data;
+    private $data;
 
     /**
      * Cookie constructor.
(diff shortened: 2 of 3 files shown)


enricodias:
Thanks for the PR.

Since the `Cookie::class` already implements `ArrayAccess`, whats the point of `getListAllDataAsArray()`?

.gitignore
```diff
@@ -3,3 +3,4 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
+.idea
```
GrahamCampbell: Please don't add this. Put this in your global ignore file.

GrahamCampbell (changes requested):

GrahamCampbell:
Can you enable editing by maintainers, so I can clean this up, please?

.gitignore
carsteneu: cant see the checkbox for "editing by maintainers"

carsteneu:
> Thanks for the PR.
> 
> Since the `Cookie::class` already implements `ArrayAccess`, whats the point of `getListAllDataAsArray()`?

I thought it might be handy... i use it in my project in this way.

GrahamCampbell:
Going to close this, since this is already possibly to implement in apps without modifying our code.

PR #318, opened 2022-01-20
[1.3] Replace microtime with hrtime

enricodias:
The system clock shouldn't interfere with the calculation of timeouts.

This PR needs to be reviewed with some care since there are no tests for the timeouts themselves.

### DIFF
--- a/src/Exception/OperationTimedOut.php
+++ b/src/Exception/OperationTimedOut.php
@@ -15,18 +15,19 @@ class OperationTimedOut extends \Exception
 {
     public static function createFromTimeout(int $timeoutMicroSec): self
     {
-        return new self('Operation timed out ('.self::getTimeoutPhrase($timeoutMicroSec).')');
+        return new self(sprintf('Operation timed out after %s.'), self::getTimeoutPhrase($timeoutMicroSec));
     }
 
     private static function getTimeoutPhrase(int $timeoutMicroSec): string
     {
         if ($timeoutMicroSec > 1000 * 1000) {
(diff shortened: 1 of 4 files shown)


src/Exception/OperationTimedOut.php
```diff
@@ -13,4 +13,20 @@
 
 class OperationTimedOut extends \Exception
 {
+    public function __construct(int $timeoutMicroSec, int $code = 0, ?\Throwable $previous = null)
```
GrahamCampbell: Increase of replacing the constructor, could you add a `public static function createFromTimeout(...)`?

GrahamCampbell (changes requested):

src/Exception/OperationTimedOut.php
enricodias: Done.

src/Exception/OperationTimedOut.php
```diff
@@ -13,4 +13,20 @@
 
 class OperationTimedOut extends \Exception
 {
+    public static function createFromTimeout(int $timeoutMicroSec): self
+    {
+        return new self('Operation timed out ('.self::getTimeoutPhrase($timeoutMicroSec).')');
+    }
+
+    private static function getTimeoutPhrase(int $timeoutMicroSec): string
```
GrahamCampbell, line 21: return type is incorrect, and only not crashing because strict types is not enabled. ;)

GrahamCampbell (changes requested):

PR #320, opened 2022-01-20
[1.3] Fix `Mouse::find()` after cursor has moved

simPod:
After scrolling to the element, its boundaries change and we have to read the viewport offsets in order to compute final coordinates.

### DIFF
--- a/src/Input/Mouse.php
+++ b/src/Input/Mouse.php
@@ -285,14 +285,16 @@ class Mouse
         $rightBoundary = \floor($element['right']);
         $bottomBoundary = \floor($element['bottom']);
 
-        $positionX = \mt_rand(\ceil($element['left']), $rightBoundary);
-        $positionY = \mt_rand(\ceil($element['top']), $bottomBoundary);
+        $this->scrollToBoundary($rightBoundary, $bottomBoundary);
 
-        $this->scrollToBoundary($rightBoundary, $bottomBoundary)
-            ->move(
-                ($positionX - $this->x),
-                ($positionY - $this->y)
-            );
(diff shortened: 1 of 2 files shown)


simPod:
this breaks `testFind_withScrolling` ummm

enricodias:
The `testFind_withScrolling` is intermittent. Is it failing every time after those changes?

simPod:
@enricodias yes but it was not working before either I think. I was just a coincidence.

simPod:
now I like it :shipit:

enricodias (approved):

src/Input/Mouse.php
```diff
@@ -285,14 +285,16 @@ public function find(string $selectors, int $position = 1): self
...
-                ($positionX - $this->x),
-                ($positionY - $this->y)
-            );
+        $this->scrollToBoundary($rightBoundary, $bottomBoundary);
+
+        $visibleArea = $this->page->getLayoutMetrics()->getLayoutViewport();
+
+        $offsetX = $visibleArea['pageX'];
+        $offsetY = $visibleArea['pageY'];
+        $positionX = \mt_rand(\ceil($element['left'] - $offsetX), $rightBoundary - $offsetX);
```
GrahamCampbell: random_int is preferable to mt_rand.

GrahamCampbell (changes requested):

src/Input/Mouse.php
simPod: I'll address it in another PR, it's not related to the PR as the functions used did not change.

PR #325, opened 2022-01-21
[1.3] Make it work with retina

simPod:
I have added a way to read css metrics that are the same as standard ones except that they consider retina displays.

Now the lib works in headful mode on retina displays.


enricodias:
🎉

PR #326, opened 2022-01-22
[1.4] Prevent `Node::querySelector` from returning nodeId 0

Seriyyy95:
Fixes bug with zero node id.

### DIFF
--- a/src/Dom/Node.php
+++ b/src/Dom/Node.php
@@ -64,7 +64,7 @@ class Node
 
         $nodeId = $response->getResultData('nodeId');
 
-        if (null !== $nodeId) {
+        if (null !== $nodeId && 0 !== $nodeId) {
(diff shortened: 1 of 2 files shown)


enricodias:
Why can't we have a node id `0`?

Seriyyy95:
querySelector returns nodeId 0 when nothing found.

enricodias:
I just tested and it really returns 0. Can you create a new test for this case? Pretty much the same as `testSearchByCssSelector`, but searching for a non-existent element and ensuring that the result is null.

https://github.com/chrome-php/chrome/blob/e9f10d0f7af33b4a58132891095c7cc525af3f11/tests/DomTest.php#L39

Seriyyy95:
Done.

enricodias (approved):

PR #334, opened 2022-01-24
[1.5] Add `waitUntilContainsElement()` to Page

p4veI:
Hello, Ive been working with the chrome lib for the past week in our project and I was unable to crawl through a js enhanced website forms without the ability to trigger some delay programmatically, unless including a long delay via `connectionDelay` option.

So I've used some parts of resolving the element in find methods and passed it through the timeout utility to wait for the element to appear with `waitUntilContainsElement`, hoping we could merge this into the main repo or would you choose a different approach?

Thx!

### DIFF
--- a/src/Page.php
+++ b/src/Page.php
@@ -472,16 +472,7 @@ class Page
         while (true) {
 
             try {
-                $elementList = $this
-                    ->evaluate('JSON.parse(JSON.stringify(document.querySelectorAll("'.$selectors.'")));')
-                    ->getReturnValue();
-
-                $position = \max(0, ($position - 1));
-                $position = \min($position, (\count($elementList) - 1));
-
-                $element = $this
-                    ->evaluate('JSON.parse(JSON.stringify(document.querySelectorAll("'.$selectors.'")['.$position.'].getBoundingClientRect()));')
-                    ->getReturnValue();
+                $element = Utils::getElementPositionFromPage($this, $selectors, $position);
             } catch (\Throwable $exception) {
                 yield $delay;
             }
--- a/src/Page.php
+++ b/src/Page.php
@@ -19,6 +19,7 @@ use HeadlessChromium\Cookies\CookiesCollection;
 use HeadlessChromium\Dom\Dom;
 use HeadlessChromium\Exception\CommunicationException;
 use HeadlessChromium\Exception\InvalidTimezoneId;
+use HeadlessChromium\Exception\JavascriptException;
 use HeadlessChromium\Exception\NoResponseAvailable;
 use HeadlessChromium\Exception\TargetDestroyed;
 use HeadlessChromium\Input\Keyboard;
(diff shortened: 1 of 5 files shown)


src/Page.php
```diff
@@ -440,6 +440,50 @@ private function waitForReloadGenerator($eventName, $loaderId)
...
+     * Wait until page contains Node.
+     *
+     * @param string $selectors
+     * @param int    $timeout
+     *
+     * @throws Exception\OperationTimedOut
+     *
+     * @return $this
+     */
+    public function waitUntilContainsElement(string $selectors, int $timeout = 30000)
```
enricodias: Can we have a return type here?

src/Page.php
```diff
@@ -440,6 +440,50 @@ private function waitForReloadGenerator($eventName, $loaderId)
...
+     */
+    public function waitForElement(string $selectors, int $position = 1)
+    {
+        $delay = 500;
+        $element = [];
+
+        while (true) {
+            try {
+                $element = Utils::getElementPositionFromPage($this, $selectors, $position);
+            } catch (\Throwable $exception) {
```
enricodias: Why catching `\Throwable` instead of exceptions? This may hide php errors.

src/Page.php
```diff
@@ -440,6 +441,48 @@ private function waitForReloadGenerator($eventName, $loaderId)
...
 
+    /**
+     * Wait until page contains Node.
+     *
+     * @param string $selectors
+     * @param int    $timeout
+     *
+     * @throws Exception\OperationTimedOut
+     */
+    public function waitUntilContainsElement(string $selectors, int $timeout = 30000) : self
```
enricodias:
```suggestion
    public function waitUntilContainsElement(string $selectors, int $timeout = 30000): self
```
To make it pass on the styleci check.

src/Page.php
p4veI: Yeah, thx, I originally experienced some other error but I resolved that later and was not relevant to the resolving here anwyays.

enricodias (approved):

src/Page.php
```diff
@@ -440,6 +441,48 @@ private function waitForReloadGenerator($eventName, $loaderId)
...
+
+        return $this;
+    }
+
+    /**
+     * @return bool|\Generator
+     *
+     * @internal
+     */
+    public function waitForElement(string $selectors, int $position = 1)
```
momala454, line 466: should probably add a timeout option to prevent infinite loot if the element never appear

src/Utils.php
```diff
@@ -90,4 +91,24 @@ public static function closeAllPage(Connection $connection): void
             }
         }
     }
+
+    /**
+     * @throws CommunicationException
+     * @throws Exception\EvaluationFailed
+     *
+     * @return mixed|null
```
GrahamCampbell: invalid return type. mixed already contains null. please remove the `|null`

GrahamCampbell (changes requested):

src/Utils.php
p4veI: ah sorry, removed. Inferred from IDE tooltip of getResult 😄

PR #336, opened 2022-01-25
[1.4] Add fallback to css layout metrics

simPod:
According to https://vanilla.aslushnikov.com/?Page.getLayoutMetrics "no-css" props are deprecated.

However, old versions of chrome might miss them.

Therefore, I'm adding fallback so it does not break old clients.

@p4vel discovered this on `Chromium 90.0.4430.212 built on Debian 10.9, running on Debian 10.11`

image


enricodias:
I'm not sure if this is a fix or a feature 🤔

simPod:
IMO fix since I might have broken something with #325

PR #338, opened 2022-01-25
[1.4] Fix missing destroyed

momala454:
Without it, it can cause
```
Call to a member function sendMessage() on null in .../Communication/Session.php:68
```

because
```php
$this->getConnection()->sendMessage
```

when it's already destroyed, destroyed is never set and then
getConnection() return null

### DIFF
--- a/src/Communication/Session.php
+++ b/src/Communication/Session.php
@@ -135,7 +135,7 @@ class Session extends EventEmitter
         }
         $this->emit('destroyed');
         $this->connection = null;
-        $this->removeAllListeners();
         $this->destroyed = true;
+        $this->removeAllListeners();
     }
 }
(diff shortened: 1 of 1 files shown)


src/Communication/Session.php
```diff
@@ -136,5 +136,6 @@ public function destroy(): void
         $this->emit('destroyed');
         $this->connection = null;
         $this->removeAllListeners();
+        $this->destroyed = true;
```
GrahamCampbell, line 138: We maybe wanna do that before calling removeAllListeners?

GrahamCampbell (changes requested):

src/Communication/Session.php
momala454: maybe but i don't know if it have any difference as removeAllListener doesn't call any other code

PR #340, opened 2022-01-31
[1.5] Add support for XPath by introducing `Selector` and `Mouse::findElement()`

### LINKED ISSUE #333: Xpath support
Hi, what would you say to adding a support for xpath to the current css selectors?

We use js already to find elements using css selectors.

Looking for elements using xpath seems to be as easy as

```js
var xpath = "//a[text()='SearchingText']";
var matchingElement = document.evaluate(xpath, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue;
```

I propose to add `Selector` interface with `CssSelector` and `XPathSelector` implementations, used as

`Mouse::find(Selector $selectors, int $position = 1): self`

in v2.0

simPod:
Resolves https://github.com/chrome-php/chrome/issues/333

### DIFF
--- a/src/Dom/Selector/Selector.php
+++ b/src/Dom/Selector/Selector.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace HeadlessChromium\Dom\Selector;
+
+interface Selector
+{
+    public function expressionCount(): string;
+
+    public function expressionFindOne(int $position): string;
+}
--- a/src/Input/Mouse.php
+++ b/src/Input/Mouse.php
@@ -12,6 +12,8 @@
 namespace HeadlessChromium\Input;
 
 use HeadlessChromium\Communication\Message;
+use HeadlessChromium\Dom\Selector\CssSelector;
+use HeadlessChromium\Dom\Selector\Selector;
 use HeadlessChromium\Exception\ElementNotFoundException;
 use HeadlessChromium\Exception\JavascriptException;
 use HeadlessChromium\Page;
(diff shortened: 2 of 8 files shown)


src/Dom/Selector/Selector.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace HeadlessChromium\Dom\Selector;
+
+interface Selector
+{
+    public function expressionCount(): string;
```
enricodias, line 9: Why use string here?

src/Input/Mouse.php
```diff
@@ -260,26 +262,54 @@ private function scrollToBoundary(int $right, int $bottom): self
...
+            $position = \max(1, $position);
+            $position = \min($position, $elementCount);
 
             $element = $this->page
-                ->evaluate('JSON.parse(JSON.stringify(document.querySelectorAll("'.$selectors.'")['.$position.'].getBoundingClientRect()));')
+                ->evaluate(\sprintf('JSON.parse(JSON.stringify(%s.getBoundingClientRect()));', $selector->expressionFindOne($position)))
                 ->getReturnValue();
         } catch (JavascriptException $exception) {
-            throw new ElementNotFoundException('The search for "'.$selectors.'" returned no elements.');
+            throw new ElementNotFoundException('The search for "'.$selector->expressionCount().'" returned no result.');
```
enricodias, line 299: The message for this exception will not mention the selector name, will it?

enricodias:
Do you think it's worth it to have a test class for each `Selector` class separately? Even thought they are still being tested indirectly in the mouse tests.

src/Dom/Selector/Selector.php
simPod: it returns an expression

src/Input/Mouse.php
simPod: https://github.com/chrome-php/chrome/pull/340#discussion_r795632223

simPod:
> Do you think it's worth it to have a test class for each Selector class separately? Even thought they are still being tested indirectly in the mouse tests.

It discovered few bugs related to position when I deved it.

simPod:
Do we want to pursue this?

enricodias:
I think so.

enricodias (approved):

simPod:
I have rebased on 1.5 since there was a force push I guess

simPod:
_The scrolling tests in GA are flaky (see https://github.com/simPod/chrome/runs/5407170225)_

enricodias:
Yes, they are. But I never managed to reproduce this error locally :/

enricodias (approved):

simPod:
> Yes, they are. But I never managed to reproduce this error locally :/

yup, same here

GrahamCampbell (changes requested):
Are you able to resolve the merge conflicts, please?

PR #342, opened 2022-02-05
[1.5] Add `Page::setHtml()`

### LINKED ISSUE #248: [Feature Request] set page HTML
I'm using this package to generate PDF from Laravel Blade templates. I can easily render the templates to HTML and then I want to generate a PDF. It would be nice if there was an easy built-in way to load HTML and not have to navigate to a URL.

I have found 2 possible ways to do this. The first one is:

```php
$page->navigate('data:text/html,'.rawurlencode($html))->waitForNavigation();
```

The second method mostly works but for some reason the image did not load but that might be a separate issue:

```php
$page->getSession()->sendMessageSync(new Message('Page.setDocumentContent', [
    'frameId' => $page->getFrameManager()->getMainFrame()->getFrameId(),
    'html' => $html,
]));
```

Maybe add a `loadHtml()` or `setHtml()` method which would be much more discoverable. It was not obvious to me how to load HTML until I figured out these tricks.

enricodias:
Resolves #248

### DIFF
--- a/src/Page.php
+++ b/src/Page.php
@@ -398,8 +398,8 @@ class Page
     {
         $this->assertNotClosed();
 
-        if (!$loaderId) {
-            $loaderId = $loader = $this->frameManager->getMainFrame()->getLatestLoaderId();
+        if (null === $loaderId) {
+            $loaderId = $this->frameManager->getMainFrame()->getLatestLoaderId();
         }
 
         Utils::tryWithTimeout($timeout * 1000, $this->waitForReloadGenerator($eventName, $loaderId));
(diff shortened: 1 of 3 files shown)


src/Page.php
```diff
@@ -384,22 +384,16 @@ public function hasLifecycleEvent(string $event): bool
...
-     * @param null   $loaderId
-     *
      * @throws CommunicationException\CannotReadResponse
      * @throws CommunicationException\InvalidResponse
      * @throws Exception\OperationTimedOut
-     *
-     * @return $this
      */
-    public function waitForReload($eventName = self::LOAD, $timeout = 30000, $loaderId = null)
+    public function waitForReload(string $eventName = self::LOAD, int $timeout = 30000, ?string $loaderId = null): self
```
GrahamCampbell: this is a major breaking change. we have to wait until 2.x to do this.

GrahamCampbell (changes requested):

GrahamCampbell (changes requested):
Merge conflicts.

GrahamCampbell (approved):
Merge after CI runs.

enricodias:
I'm not sure why the CI tests are not being triggered. 🤔

GrahamCampbell:
Because there is an actions outage. Been down basically all week.

GrahamCampbell:
It'll run eventually. Backlog is a few hours, I think.

PR #345, opened 2022-02-18
[1.5] Fix cookie expiration example

maldoinc:
Fix the given example for the cookie expiration time as `time() + 3600` refers to the time in one hour rather than in one day from now


enricodias:
Thank you.

GrahamCampbell:
This is a bug fix, so should go to 1.4. :)

enricodias:
I moved the commit to 1.4 as well.

PR #351, opened 2022-03-02
[1.5] Added `getPages` function to Browser class

edgardmessias:
Added getPages function to Browser class.


enricodias:
The intent is to be able to have several pages opened at the same time, right? If that's a new feature, it would be nice to add it to the docs.

edgardmessias:
I made some changes know, let me know if you want more changes or suggestions

enricodias (approved):

PR #352, opened 2022-03-02
Added compatibility for composer >= 2.2.0

edgardmessias:
Added compatibility for composer >= 2.2.0

Ref: https://getcomposer.org/doc/06-config.md#allow-plugins

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,6 +41,9 @@
         }
     },
     "config": {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, since this is a bug fix, I've applied it to 1.4 branch.

PR #356, opened 2022-03-18
[1.4] Change `uploadFile()` to `sendFile()` in README.md

ltdangle:
Current upload file example is:
`$elem->uploadFile('/path/to/file');`
while it should be:
`$elem->sendFile('/path/to/file');`


enricodias:
Thank you.

PR #358, opened 2022-03-24
[1.4] Fix typo in README.md

JoolsMcFly:
Hi guys,

just tried chrome-php today and found a typo while reading the doc: `tempalte` instead of `template` so here's my modest contribution!


enricodias:
Thank you :)

PR #382, opened 2022-05-21
[1.8] Better autocomplete in search function

FelipeAzambuja:
Most IDE use phpdoc for autocomplete helper , its make more easy to code


enricodias:
Thanks for the contribution. I think that the recommended syntax is `@return \HeadlessChromium\Dom\Node[]`, according the the [documentation](https://docs.phpdoc.org/latest/guide/references/phpdoc/types.html#arrays).
Also, can you fix the styling issues pointed by StyleCI?

PR #390, opened 2022-06-17
Setting Extra HTTP Headers Not Working

khoubeibouthour:
## Issue confirmed and tested only with Google Chrome
Adding option `headers` to the constructor is returning an error in the log `{"code":-32601,"message":"'Network.setExtraHTTPHeaders' wasn't found"}`.
When calling `setEtrxaHTTPHeaders` on the page, the method is accepted but it complains that the format is incorrect `Failed to deserialize params.headers`
It looks like Chrome API expects the headers array to be nested one level deeper under key `headers`.

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -82,10 +82,6 @@ class Browser
 
         // enable target discovery
         $connection->sendMessageSync(new Message('Target.setDiscoverTargets', ['discover' => true]));
-
-        // set up http headers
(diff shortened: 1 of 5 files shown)


enricodias:
Removing those public methods will break backwards compatibility.

theonlymeliii:
`gh pr checkout 390`

PR #407, opened 2022-08-20
[1.8] Add enum-like class `BrowserOptions` with all options for the browser creation

### LINKED ISSUE #332: Update formatting of options for readability
https://github.com/chrome-php/chrome/blob/1.3/src/BrowserFactory.php#L29-L48

```php
/**
 * Options for browser creation.
 *
 * - connectionDelay:           Delay to apply between each operation for debugging purposes (default: none)
 * - customFlags:               An array of flags to pass to the command line.
 * - debugLogger:               A string (e.g "php://stdout"), or resource, or PSR-3 logger instance to print debug messages (default: none)
 * - enableImages:              Toggles loading of images (default: true)
 * - envVariables:              An array of environment variables to pass to the process (example DISPLAY variable)
 * - headers:                   An array of custom HTTP headers
 * - headless:                  Enable or disable headless mode (default: true)
 * - ignoreCertificateErrors:   Set chrome to ignore ssl errors
 * - keepAlive:                 Set to `true` to keep alive the chrome instance when the script terminates (default: false)
 * - noSandbox:                 Enable no sandbox mode, useful to run in a docker container (default: false)
 * - proxyServer:               Proxy server to use. ex: `127.0.0.1:8080` (default: none)
 * - sendSyncDefaultTimeout:    Default timeout (ms) for sending sync messages (default 5000 ms)
 * - startupTimeout:            Maximum time in seconds to wait for chrome to start (default: 30 sec)
 * - userAgent:                 User agent to use for the whole browser
 * - userDataDir:               Chrome user data dir (default: a new empty dir is generated temporarily)
 * - windowSize:                Size of the window. ex: `[1920, 1080]` (default: none)
 */
```

enricodias:
It's easier to use and we can rely on the IDE's auto completion feature.
In future versions, it can be turned into a real `Enum` without breaking backwards compatibility since the usage syntax is the same.

I'm not sure if this is the best approach, but it's better than a huge docblock listing all the options. I'm open to better ideas.

Resolves #332

### DIFF
--- a/src/Browser/BrowserOptions.php
+++ b/src/Browser/BrowserOptions.php
@@ -1,0 +1,115 @@
+<?php
+
+/*
+ * This file is part of Chrome PHP.
+ *
+ * (c) Soufiane Ghzal <sghzal@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace HeadlessChromium\Browser;
+
+abstract class BrowserOptions
+{
+    /**
+     * Delay to apply between each operation for debugging purposes.
+     */
+    public const connectionDelay = 'connectionDelay';
+
+    /**
+     * An array of flags to pass to the command line.
+     */
+    public const customFlags = 'customFlags';
+
+    /**
+     * A string (e.g "php://stdout"), or resource, or PSR-3 logger instance to print debug messages.
+     */
+    public const debugLogger = 'debugLogger';
+
+    /**
+     * Toggles loading of images.
+     *
+     * Default: true
+     */
+    public const enableImages = 'enableImages';
+
+    /**
+     * An array of environment variables to pass to the process.
+     *
+     * Example: DISPLAY
+     */
+    public const envVariables = 'envVariables';
+
+    /**
+     * An array of custom HTTP headers.
+     */
+    public const headers = 'headers';
+
+    /**
+     * Enable or disable headless mode.
+     *
+     * Default: true
+     */
+    public const headless = 'headless';
+
+    /**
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
This is not a bug fix, so would need to target 1.7. :)

enricodias:
I forgot to change the version. I'll fix the merge conflict and change the target of this and the other features to 1.7.

But are you ok with this approach?

src/Browser/BrowserOptions.php
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * (c) Soufiane Ghzal <sghzal@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace HeadlessChromium\Browser;
+
+abstract class BrowserOptions
```
GrahamCampbell, line 14: I actually don't like having a separate class for this. Best practice is to have constants with a specific prefix on the original interface/class itself. See https://psalm.dev/docs/annotating_code/typing_in_psalm/#specifying-stringint-options-aka-enums.

GrahamCampbell (changes requested):

src/Browser/BrowserOptions.php
enricodias: I was thinking that when this class gets converted to an enum in the future, we could restrict the argument type in the `createBrowser` method to this enum only. Using something like `[BrowserFactory::OPTIONS_headless => true]` looks more verbose and a bit ugly to me. 🤔

GrahamCampbell:
I don't think this is necessary. Maybe in the next major version we can bump to to PHP 8.1 as the minimum and use enums.

PR #408, opened 2022-08-20
[1.8] Add enum-like class `PagePdfOptions` with all options for pdf creation

enricodias:
Same motivation as #407


GrahamCampbell:
This is not a bug fix, so would need to target 1.7. :)

enricodias:
Not stale.

GrahamCampbell:
I don't think this is necessary. Maybe in the next major version we can bump to to PHP 8.1 as the minimum and use enums.

PR #413, opened 2022-08-23
[1.7] Fix extra http headers

### LINKED ISSUE #404: Not working custom headers
```
require_once 'vendor/autoload.php';

    'windowSize' => [1920, 1080],
    'enableImages' => true,
    'headless' => false,
    'headers' => [
        'Accept: */*',
        'sec-fetch-dest: document',
        //'test' => 'document', // or this
    ]
]);
$page->navigate('http://foxtools.ru/Http');
sleep(20);
$browser->close();

exit();
```

enricodias:
The correct syntax is `['headers' => $headers]`.
`Network.setHTTPHeaders` must be sent per page instance, after `Network.enable`.

`Page::setExtraHTTPHeaders` will now throw an exception if the response contains an error. The tests were modified to actually try to send the headers to chrome, to make them fail if we get an exception.

Fixes  #404

### DIFF
--- a/src/Browser.php
+++ b/src/Browser.php
@@ -182,9 +182,13 @@ class Browser
     }
 
     /**
+     * @param string $targetId
+     *
      * @throws CommunicationException
+     *
+     * @return Page|null
      */
-    public function getPage(string $targetId): ?Page
+    public function getPage($targetId)
     {
         if (\array_key_exists($targetId, $this->pages)) {
             return $this->pages[$targetId];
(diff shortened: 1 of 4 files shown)


src/Browser.php
```diff
@@ -186,11 +182,9 @@ public function getTargets()
     }
 
     /**
-     * @param string $targetId
-     *
-     * @return Page|null
+     * @throws CommunicationException
      */
-    public function getPage($targetId)
+    public function getPage(string $targetId): ?Page
```
GrahamCampbell: this is a major breaking change. we can't add param types to public methods, and can't add return types to public or protected methods on non-final classes

GrahamCampbell (changes requested):

PR #414, opened 2022-08-23
Add lang option for browser creation

otsch:
You can set the browser language to use via the --lang command line option. Added it to the possible options for browser creation.

I think it should also be possible via custom flags, but I think there could be a lot of users needing it that don't know the command line flag.

Just found that there is already a 1.7 branch before pushing, so I made it the target branch.
I didn't create an issue first because it was implemented pretty fast. So, no big problem if you don't want to merge it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # CHANGELOG
 
 
+## 1.7.0 (2022-08-24)
+
+* Added option `lang` to browser creation.
+
+
 ## 1.6.1 (2022-05-17)
 
 * Support Monolog 3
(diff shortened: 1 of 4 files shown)


enricodias:
Thanks for contributing :)

Your new test doesn't seem to pass in the CI tests. Does it work in your local env?

otsch:
Yes, it works on my machine (macOS). But on the other hand there are other tests that fail in my local env. I'll try to set up a testing environment in docker today and try again.

otsch:
OK, after a lot of research, I think this flag just doesn't work with chromium on linux. I found advice to set the language via environment variable when starting the chrome browser, but didn't work for me.

As I've seen that there's [another open PR](https://github.com/chrome-php/chrome/pull/413) by you @enricodias, to fix sending HTTP headers, one can use the `accept-language` HTTP header instead of this. So, I'll just close this PR.

PR #424, opened 2022-09-03
[1.7] Fix command execution in `AutoDiscover`

enricodias:
`command -v` was being executed in the call inside `getFromRegistry()`.
Tests were failing when running without the `CHROME_PATH` env var.

### DIFF
--- a/tests/AutoDiscoverTest.php
+++ b/tests/AutoDiscoverTest.php
@@ -22,7 +22,7 @@ class AutoDiscoverTest extends BaseTestCase
 
     protected function setUp(): void
     {
-        if (false !== \array_key_exists('CHROME_PATH', $_SERVER)) {
+        if (\array_key_exists('CHROME_PATH', $_SERVER)) {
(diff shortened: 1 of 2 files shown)


tests/AutoDiscoverTest.php
```diff
@@ -22,9 +22,11 @@ class AutoDiscoverTest extends BaseTestCase
 
     protected function setUp(): void
     {
-        $this->originalEnvPath = null ?? $_SERVER['CHROME_PATH'];
+        if (false !== \array_key_exists('CHROME_PATH', $_SERVER)) {
+            $this->originalEnvPath = null ?? $_SERVER['CHROME_PATH'];
```
GrahamCampbell: `null ??` doesn't do anything, and can be removed.

GrahamCampbell (changes requested):

tests/AutoDiscoverTest.php
enricodias: Fixed.

PR #430, opened 2022-09-10
[1.8] Add `NetworkResponseEntity::class` to store network responses

### LINKED ISSUE #41: Get response status code
I need to get response status code, what can I do?

### LINKED ISSUE #434: How to get all Network conditions like DevTools
Hi.

Can I get all requests conditions like "Network" in Chrome DevTools?
All means actual js or json in script tags, js or json called in js.
 
I'm going to crawl a specific API calling in the pages.

enricodias:
All network responses since the last navigation will be saved as `NetworkResponseEntity`. They can be fetched using  `Page::getNetworkResponses()`.

At first, my intention was to save just the data of the html request, but then we would lose data in case of redirects. It's reasonable to assume that data from all redirects can be important, even redirects triggered by javascript after the page was loaded. So, the final feature just stores all network responses.

Resolves #41
Resolves #434


enricodias:
Maybe it will look nicer without the "Entity" in the name, just `NetworkResponse` 🤔

src/Entity/NetworkResponseEntity.php
```diff
@@ -0,0 +1,79 @@
...
+    ) {
+        $this->url = $url;
+        $this->status = $status;
+        $this->statusText = $statusText;
+        $this->headers = $headers;
+        $this->mimeType = $mimeType;
+    }
+
+    /**
+     * Readonly trick for php < 8.1.
```
GrahamCampbell, line 71: I don't think this will work. __set only fires when there's not an actual property here. See https://3v4l.org/LVp0F. For this reason, I feel like either these should be replaced with private properties and getter methods, or alternatively, we should not bother with an entity class at all, and just use typed arrays.

```php
    /**
     * Network responses from all requests since the last navigation.
     *
     * @var array{url: string, status: int, statusText: string, headers: array<string, string>, mimeType: string}[]
     */
    private $networkResponses = [];
```

GrahamCampbell (changes requested):

src/Entity/NetworkResponseEntity.php
enricodias: How about using `__get` with private properties and add `@property-read` tags in the class docblock? https://3v4l.org/nFdbN

The types will be enforced, IDEs will show the properties in the auto complete feature, those properties will be read only and wel'll be able to replace everyting with the `readonly` keyword in php >=8.1 without changing the class behaviour.

PR #435, opened 2022-09-14
[1.7] Add windows compatibility for unit tests

ttk:
Prior to this change, 2 unit tests were failing.  This change allows for all unit tests to pass on Windows.

### DIFF
--- a/tests/BaseTestCase.php
+++ b/tests/BaseTestCase.php
@@ -17,6 +17,11 @@ class BaseTestCase extends TestCase
 {
     protected static function sitePath(string $file): string
     {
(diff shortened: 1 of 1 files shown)


enricodias:
I tested it in a windows machine and it works. Thank you.

GrahamCampbell:
We should really be running CI on Windows, if we want to commit to Windows support?

GrahamCampbell:

enricodias:
I'm not sure if adding CI on Windows is worth it. The lib itself should work on windows anyway, most of the merges related to windows support are just annoyances in the test suite. I'm accepting this kind of PR because seen tests failing like this may be a bummer for new people looking to contribute.

PR #438, opened 2022-09-17
[1.7] Fix timezone test failing in non english locales

enricodias:
The end of the `date.toString()` gets translated to the local language.


GrahamCampbell:

PR #440, opened 2022-10-07
[1.7] Fix hard coded timeout in `Dom` class

### LINKED ISSUE #395: How to increase timeout in `Page::dom()`
I get operation timeout error 
is there anyway i can increase time it takes to load dom.

enricodias:
This will make the `Dom` constructor use the `sendMessageSync()` method like the rest of the class, which uses the timeout set by the option `sendSyncDefaultTimeout` when creating a browser instance with the `BrowserFactory` class.

Fixes #395

### DIFF
--- a/src/Dom/Dom.php
+++ b/src/Dom/Dom.php
@@ -12,8 +12,7 @@ class Dom extends Node
     public function __construct(Page $page)
     {
         $message = new Message('DOM.getDocument');
-        $stream = $page->getSession()->sendMessage($message);
-        $response = $stream->waitForResponse(1000);
+        $response = $page->getSession()->sendMessageSync($message);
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Doesn't this mean there is now no timeout?

enricodias:
If `sendSyncDefaultTimeout` is not set, 5 seconds is used by default: https://github.com/chrome-php/chrome/blob/a67e5005f09e71e637ff04839d822b8e47aee545/src/Communication/Connection.php#L94

It's also documented in the readme: https://github.com/chrome-php/chrome/blob/113850cfccfdda6d7200d0466e6315dc4b83f74c/README.md?plain=1#L175

GrahamCampbell:

PR #455, opened 2022-12-14
Wait for the NETWORK_IDLE event instead of LOAD when setting a page's HTML through the setHtml method

sanderdlm:
I've been playing with this locally for a while, and using NETWORK_IDLE produces far more consistent results when using external CSS/JS in your PDF templates. For example, loading a Google Font through their CDN would sometimes produce a PDF where the font would be missing, but with networkIdle, the font is always there.

If we can't change the default, maybe we should look into passing the event to wait for as an option? That way we can keep LOAD as the default but give us the tools to change it.

Thanks!

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+CHROME_PATH=/usr/bin/chromium
--- a/src/Page.php
+++ b/src/Page.php
@@ -874,7 +874,7 @@ class Page
 
         $timeout -= (int) \floor((\hrtime(true) / 1000 / 1000) - $time);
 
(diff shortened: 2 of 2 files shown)


enricodias:
I see your point, but the network may never be idle in some cases. If you have a js making infinite requests in the background or the backend extending response times intentionally (long pooling), you'll get a timeout exception. Maybe this could be an optional argument so it doesn't change the behavior and break previous implementations.

Also, you committed your env file.

sanderdlm:
I get your point. Moved this over to a new PR with an optional parameter: https://github.com/chrome-php/chrome/pull/456

PR #466, opened 2023-01-09
[1.7] Stop process when an error occurs during startup

### LINKED ISSUE #465: Does not close browser when an error occur during startup
If we set the port to 9222 using `--remote-debugging-port=9222` by example, while we set `keepAlive` to true using but this port is already used, we don't get a web socket address, the process is kept open without the possibility to close it because we don't have any reference to the buggy process.

Chrome will use ipv4 on the first instance, then ipv6 as it detect the port is already used on ipv4, and on the third process, ipv4 and ipv6 are used, so an exception is thrown (HeadlessChromium\\Exception\\OperationTimedOut because the output "DevTools listening on" wasn't found

To reproduce

```php
$browserFactory = new \HeadlessChromium\BrowserFactory('C:/Program Files/Google/Chrome/Application/chrome.exe');

  'userAgent'					=>	'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.99 Safari/537.36',
  'headless'                    =>  false,
  //'debugLogger'     			=> 'php://stdout',
  'customFlags'                 => ['--no-sandbox', '--remote-debugging-port=9222'],
  'keepAlive'                   =>  true,
]);

echo $browser->getSocketUri()."\n";

    'userAgent'					=>	'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.99 Safari/537.36',
    'headless'                    =>  false,
    //'debugLogger'     			=> 'php://stdout',
    'customFlags'                 => ['--no-sandbox', '--remote-debugging-port=9222'],
    'keepAlive'                   =>  true,
  ]);
  echo $browser->getSocketUri()."\n";

    'userAgent'					=>	'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/97.0.4692.99 Safari/537.36',
    'headless'                    =>  false,
    'debugLogger'     			=> 'php://stdout',
    'customFlags'                 => ['--no-sandbox', '--remote-debugging-port=9222'],
  ]);

  echo $browser->getSocketUri()."\n";

 ```

momala454:
Fixes https://github.com/chrome-php/chrome/issues/465

### DIFF
--- a/src/Browser/BrowserProcess.php
+++ b/src/Browser/BrowserProcess.php
@@ -446,6 +446,7 @@ class BrowserProcess implements LoggerAwareInterface
 
             return Utils::tryWithTimeout($timeout, $generator($process));
         } catch (OperationTimedOut $e) {
+            $process->stop();
             throw new \RuntimeException('Cannot start browser', 0, $e);
         }
     }


momala454:
we could also listen for the text "Cannot start http server for devtools."
to not wait for the timeout to close the process

GrahamCampbell:

PR #468, opened 2023-01-12
[1.7] Disconnect socket on sendCloseMessage

### LINKED ISSUE #467: Closing the browser cause a timeout after
Disconecting  the socket and preventing sending "Browser.close" if the socket is not connected fixed the problem

momala454:
Also prevent sending close message again, if already disconnected

fixes https://github.com/chrome-php/chrome/issues/467


GrahamCampbell:

PR #471, opened 2023-01-19
[1.7] Stop browser on waitForStartup if devTools failed to start

momala454:
Because we will never receive "DevTools listening on", so there is no point to continue waiting


GrahamCampbell:

PR #472, opened 2023-01-27
[1.7] Actually close browser when devtools cannot start

momala454:
Instead of just throwing exception (`catch OperationTimeout` does not catch `RuntimeException`)


GrahamCampbell:

PR #483, opened 2023-02-27
[Docs]: Fix default paperHeight value in the comment

kingjia90:
See https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF:~:text=Paper%20height%20in%20inches.%20Defaults%20to%2011%20inches.


enricodias:
Thanks.

PR #501, opened 2023-03-21
[1.7] Fixed issue where Chrome 111 and later have different default allowed origins

dbisso:
For situations where PHP 7.3 is still required, it would be useful to support the newest version of Chrome which requires a fix for the changes to the allow remote origin argument

This PR just back-ports https://github.com/chrome-php/chrome/pull/497 to the 1.7 branch, which still supports PHP 7.3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fixed issue where Chrome 111 and later have different default allowed origins
 
+
 ## 1.7.2 (2023-02-27)
 
 * Fix case where the timeout in `setHtml()` isn't respected
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## 1.7.3 (2023-03-XX)
+
+* Fixed issue where Chrome 111 and later have different default allowed origins
 
 ## 1.7.2 (2023-02-27)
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
PHP 7.3 has been EOL for more than a year now, and we are only supporting the latest 1.x line of this package.

GrahamCampbell:
If you need to use EOL PHP, I guess you might as well use EOL Chrome too. ;)

momala454:
Hello @GrahamCampbell you're also blocking on php <7.4.15

`- chrome-php/chrome v1.8.1 requires php ^7.4.15 || ^8.0.2 -> your PHP version (7.4.3) does not satisfy that requirement.`

7.4.3 is the version on debian stable and ubuntu 20 LTS. I know that it's possible to to install apt depots to install an higher version.
Just saying.

GrahamCampbell:
PHP less than that version is also EOL, with lots of known security issues and bugs.

GrahamCampbell:
If people are using a patched version of an 7.4.3 with security backports, they can always set a PHP version in their app composer.json to make composer think it's a more recent version of 7.4.x. ;)

GrahamCampbell:
For example:

```
    "config": {
        "platform": {
            "php": "7.4.16"
        }
    }
```

GrahamCampbell:
[CVE-2020-7071](https://www.cvedetails.com/cve/CVE-2020-7071/) in PHP 7.4.14 is actually relevant for us, as it relates to URL validation.

PR #503, opened 2023-03-28
Add Page.handleJavaScriptDialog method

gdlapka:
Adding ability to send Page.handleJavaScriptDialog to interaction with JS alerts via Page.javascriptDialogOpening event

### DIFF
--- a/src/Page.php
+++ b/src/Page.php
@@ -1056,28 +1056,23 @@ class Page
     /**
      * Handle JS dialog
      *
-     * @param array $options
-     *
+     * @param bool $isAccept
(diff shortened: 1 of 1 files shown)


enricodias:
Thanks for the contribution.

The method seems useful, but we need a test for it and fix the styleci issues. Also, if the option is boolean in the api, there's no reason for using an array as an argument, we can use a `bool`. The method is also missing a return type.

GrahamCampbell:
Closing due to inactivity. Please feel free to re-submit with the above feedback addressed.

PR #506, opened 2023-04-07
Handling java script dialog

gdlapka:


gdlapka:
Thanks for answering. #506 opened

PR #509, opened 2023-04-20
[1.9] Added some lifecycle events

buismaarten:
When I log the results of the `getCurrentLifecycle` method, it shows lifecycle events for which there is not yet a constant inside the `Page` class. Constants are sorted alphabetically.

Reference: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/http/tests/inspector-protocol/page/page-lifecycleEvents-expected.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,12 @@ $navigation->waitForNavigation(Page::DOM_CONTENT_LOADED, 10000);
 Available events (in the order they trigger):
 
 - ``Page::DOM_CONTENT_LOADED``: dom has completely loaded
+- ``Page::FIRST_CONTENTFUL_PAINT``: triggered when the first non-white content element is painted on the screen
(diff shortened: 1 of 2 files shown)


buismaarten:
@GrahamCampbell ping

enricodias:
If we are gonna map new events, the [doc](https://github.com/chrome-php/chrome/blob/9146c37c14d99cff3e4d9cd2f5ba843d149b3cbc/README.md?plain=1#L266) should be updated as well.

buismaarten:
> If we are gonna map new events, the [doc](https://github.com/chrome-php/chrome/blob/9146c37c14d99cff3e4d9cd2f5ba843d149b3cbc/README.md?plain=1#L266) should be updated as well.

I just updated the documentation, could you have a look?

GrahamCampbell:
Thanks guys. 🎉

PR #513, opened 2023-04-26
Use `--disable-notifications` flag to disable notifications

buismaarten:


buismaarten:
Some tests failed unexpectedly, can you run again?

enricodias:
Adding that way will break backwards compatibility as there may be people parsing notifications with this lib. Making it customizable with the `$options` argument is safer.

buismaarten:
Yeah, I didn't think of that.. I'll make the changes to soon.

enricodias:
And since it will be a new feature, it should be in v1.9 too.

PR #514, opened 2023-04-26
Added `--disable-ipc-flooding-protection` flag

buismaarten:


enricodias:
I believe that this one is safe to add by default, but since it's not a fix we should target the 1.9 branch.

buismaarten:
I'll make a PR to the 1.9 branch soon.

momala454:
I don't get why it should be disabled?

PR #526, opened 2023-05-26
[1.9] Use RuntimeException in JavascriptException

buismaarten:
Hey there,

I want to semantically separate the exceptions that can be thrown by this package.

When the JavascriptException is thrown, it means that something happened during the execution of the given JavaScript code. If JavascriptException extends RuntimeException, it can be caught by the same try-catch statement to catch other application-specific errors that throws RuntimeException.

JavascriptException currently extends the Exception class, RuntimeException also extends the Exception class. No breaking changes here.

### DIFF
--- a/src/Exception/JavascriptException.php
+++ b/src/Exception/JavascriptException.php
@@ -11,6 +11,6 @@
 
 namespace HeadlessChromium\Exception;
 
-class JavascriptException extends \Exception
+class JavascriptException extends \RuntimeException
 {
 }
--- a/src/Page.php
+++ b/src/Page.php
@@ -887,7 +887,6 @@ class Page
      * Gets the raw html of the current page.
      *
      * @throws CommunicationException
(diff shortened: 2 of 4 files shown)


GrahamCampbell (changes requested):
If we do this, we'd also need to remove `JavascriptException` from all throws blocks. Every PHP method already has an implicit throws `RuntimeException|Error`. That is, throws doc is only for documenting non-runtime exceptions and non-errors.

buismaarten:
Made some changes, hopefully I made the right changes!

buismaarten:
@GrahamCampbell ping

GrahamCampbell:
I think this is probably a breaking change, so this may need to wait till 2.x.

GrahamCampbell:
Breaking because someone catching `RuntimeException` now catches this when they weren't expecting to.

PR #528, opened 2023-05-31
[1.9] Add argument for default value to `getResultData` method of `Response` class

buismaarten:


GrahamCampbell:
I typically don't like these kind of adjustments, as they break separation of concerns. One should just put the `??` on the end when you call the function.

PR #529, opened 2023-06-06
[1.9] Added return types

buismaarten:


GrahamCampbell:
Thanks but these are breaking changes to non-final classes. We could consider this for the next major version, though.

PR #531, opened 2023-06-10
[1.10] Add waitUntilContainsXPathElement method

buismaarten:


buismaarten:
@GrahamCampbell ping

enricodias:
It seems like the same logic as `waitForElement`, but using `XPathSelector` instead of `CssSelector`.

I don't like removing types, but it seems better to change `$selectors` a mixed type and let it receive a type `string`, `CssSelector` or `XPathSelector`. We could also add a third optional argument as a flag to specify what kind of selector we want to use. I'm not sure which one is better.

PR #534, opened 2023-06-29
[1.10] Allow to remove default chrome flags via option excludedSwitches

momala454:
Example, to remove the bar "Chrome is being controlled ...", you must not send the flag `--enable-automation`, so by adding this flag to the excludedSwitches, the bar is absent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ Here are the options available for the browser factory:
 | `userDataDir`             | none    | Chrome user data dir (default: a new empty dir is generated temporarily)                     |
 | `userCrashDumpsDir`       | none    | The directory crashpad should store dumps in (crash reporter will be enabled automatically)  |
 | `windowSize`              | none    | Size of the window. usage: `$width, $height` - see also Page::setViewport                    |
+| `excludedSwitches`        | none    | Chrome flags that should be removed from the default set (example --enable-automation)       |
 
 
 ### Persistent Browser
--- a/README.md
+++ b/README.md
@@ -157,29 +157,29 @@ $browser4 = $browserFactory->createBrowser();
 
 Here are the options available for the browser factory:
 
-| Option name               | Default | Description                                                                                  |
-|---------------------------|---------|----------------------------------------------------------------------------------------------|
-| `connectionDelay`         | `0`     | Delay to apply between each operation for debugging purposes                                 |
-| `customFlags`             | none    | An array of flags to pass to the command line. Eg: `['--option1', '--option2=someValue']`    |
-| `debugLogger`             | `null`  | A string (e.g "php://stdout"), or resource, or PSR-3 logger instance to print debug messages |
-| `disableNotifications`    | `false` | Disable browser notifications                                                                |
-| `enableImages`            | `true`  | Toggles loading of images                                                                    |
-| `envVariables`            | none    | An array of environment variables to pass to the process (example DISPLAY variable)          |
-| `headers`                 | none    | An array of custom HTTP headers                                                              |
-| `headless`                | `true`  | Enable or disable headless mode                                                              |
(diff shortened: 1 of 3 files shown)


enricodias:
The reason you don't just do `$args = \array_diff($args, $options['excludedSwitches']);` is because there are some options with values that wouldn't match, right? If we do something like this we'll have the same effect but without needing to put the extra `--`:

```php
if (\array_key_exists('excludedSwitches', $options) && \is_array($options['excludedSwitches'])) {
    $excludedArgs = \implode(',', $options['excludedSwitches']);
    $args = \array_filter($args, function ($arg) use ($excludedArgs) {
        return \stripos($arg, $excludedArgs) === false;
    });
}
```

But that look awful.

Can't we just remove the options that have no value using `array_diff`? The ones that have value the user can override with his own value, even an empty value. Actually, passing the value with the switch in the `excludedSwitches` will work too, since it will work as a generic filter.

There's also a docblock describing those options that would need to be updated here: https://github.com/chrome-php/chrome/blob/ab591afbfd973e1950a07c166d2f60f7e13033d7/src/BrowserFactory.php#L28-L50

The docs also need to state that this option expects an array, not just a string.

And btw, I tested here and chrome still shows the message that chrome is being controlled, even without the switch `--enable-automation`.

momala454:
i'm not using array_diff to not have to put exactly the full parameter, if we have by example
`--user-data-dir=/home/hello`

having as excluded `--user-data-dir` without having to write the parameter value will work.
stripos could cause false positive if you exclude --user but you have a parameter --user-data-dir by example

I have tested on windows and I also modified another preference (CommandLineFlagSecurityWarningsEnabled to false) via a policy (https://sites.google.com/a/chromium.org/dev/administrators/windows-quick-start). The policy itself without the enable-automation removing didn't worked, maybe it's the combination of the two changes that made it work,


some parameters like proxy-server, user-agent, --remote-debuguing-port are varying, so having to put exactly the value that was provided could be a bit difficult.
Maybe remove the excludedSwitches before adding the custom flags so we can remove the default one and put our custom ones

enricodias:
> stripos could cause false positive if you exclude --user but you have a parameter --user-data-dir by example

Good catch, I missed that case.

The `user-data-dir` is the only one we can't filter out using `array_diff` because we don't know its default value, right? But if we just filter it like you are doing, a temp folder will still be created and removed after each execution: https://github.com/chrome-php/chrome/blob/ab591afbfd973e1950a07c166d2f60f7e13033d7/src/Browser/BrowserProcess.php#L107-L109
If the only reason for filtering the user dir is to let chrome use its default profile, maybe we should just have a method for that and ignore all things related to the user dir entirely.

All the other options with values can be overwritten and the ones without value can be filtered out with a simple diff.

momala454:
User data dit was just an example. User agent is also configured elsewhere, the values of some default parameters could also be modified between lib versions, so one could want to be able to remove parameters regardless of the value.

enricodias:
But we can, by overwriting them with our own values. The lib uses the default values only when the option isn't defined, except for those first ones like the `enable-automation` in your example, that could be filtered with a simple diff. The first 2 options on that list have values (remote-debugging-port and remote-allow-origins), but we wouldn't be able to control chrome without them.

The `user-data-dir` is special because it's used elsewhere. If we're gonna let the user not use any dir, we should not create and remove a temp folder in those cases. So this would require more than a simple filter.

We can't let the user filter values that will prevent the lib from controlling chrome. It will timeout and I think that chrome process could remain alive after the php process dies.

The recent changes in the `getHtml` are not related to this feature.

enricodias (approved):

momala454:
@GrahamCampbell can this be merged please ? thanks

GrahamCampbell:
Thanks.

PR #541, opened 2023-08-28
[1.10] Add `setScriptExecution` page method

osbre:
Same as https://github.com/puppeteer/puppeteer/blob/c4bad4a6dade2eafa6a8145077a1103153467e8e/packages/puppeteer-core/src/common/EmulationManager.ts#L212-L220

### DIFF
--- a/src/Page.php
+++ b/src/Page.php
@@ -1074,9 +1074,9 @@ class Page
      *
      * @param bool $enabled
      *
-     * @return ResponseWaiter
-     *
      * @throws CommunicationException
+     *
+     * @return ResponseWaiter
      */
     public function setJavaScriptEnabled(bool $enabled)
     {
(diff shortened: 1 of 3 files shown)


enricodias:
Thanks for your contribution. I think it's a useful feature, but it's kinda odd to call a function with "enabled" in the name to disable something. How about just `setScriptExecution(bool $enabled)`?

osbre:
Thanks, @enricodias. Updated the name.

src/Page.php
```diff
@@ -1068,4 +1068,25 @@ public function setUserAgent(string $userAgent)
...
+    /**
+     * Disable or enable JavaScript execution for the current page.
+     *
+     * @param bool $enabled
+     *
+     * @throws CommunicationException
+     *
+     * @return ResponseWaiter
+     */
+    public function setScriptExecution(bool $enabled)
```
enricodias:
```suggestion
     * @throws CommunicationException
     */
    public function setScriptExecution(bool $enabled): ResponseWaiter
```

Those param and return docblocks are unnecessary if we use the types correctly directly in the method itself. Other then that, the code looks good to me.

GrahamCampbell:
Note that we follow semver, so this needs to target the 1.10 branch.

enricodias (approved):

GrahamCampbell:
Thanks.

PR #544, opened 2023-09-06
[1.10] Allow usage of XPath in waitForElement

### LINKED ISSUE #539: Allow usage of XPathSelector in Page::waitForElement
https://github.com/chrome-php/chrome/blob/f9fcb66e5c5adc11900f93fde5b002de8aa10bfb/src/Page.php#L483

https://github.com/chrome-php/chrome/pull/531#issuecomment-1587824411

It's always using CssSelector.  

It should be possible to pass an instance of a custom selector.

```

   /**
     * Wait until page contains Node.
     *
     * @param string|Selector $selectors
     * @param int $timeout = 30000
     * @throws Exception\OperationTimedOut
     */
    public function waitUntilContainsElement($selectors, int $timeout = 30000): self
    {
        $this->assertNotClosed();

        Utils::tryWithTimeout($timeout * 1000, $this->waitForElement($selectors));

        return $this;
    }

    /**
     * @param string|Selector $selectors
     * @param int $position = 1
     * @return bool|\Generator
     *
     * @internal
     */
    public function waitForElement($selectors, int $position = 1)
    {
        if (!($selectors instanceof Selector)) {
            $selectors = new CssSelector($selectors);
        }

        $delay = 500;
        $element = [];

        while (true) {
            try {
                $element = Utils::getElementPositionFromPage($this, $selectors, $position);
            } catch (JavascriptException $exception) {
                yield $delay;
            }

            if (\array_key_exists('x', $element)) {
                return true;
            }

            yield $delay;
        }
    }

jhoopmann:
https://github.com/chrome-php/chrome/issues/539

### DIFF
--- a/README.md
+++ b/README.md
@@ -791,7 +791,7 @@ $page->waitUntilContainsElement('div[data-name=\"el\"]');
 ```
 
 If a string is passed to `Page::waitUntilContainsElement`,
-an instance of CSSSelector is created by default for internal `Page::waitForElement`.
+an instance of CSSSelector is created by default in called internal `Page::waitForElement`.
 To use other selectors, you can pass an instance of the required Selector instead of a string.
 
 Wait for element by XPath selector:
--- a/README.md
+++ b/README.md
@@ -790,9 +790,9 @@ $page->navigate('http://example.com')->waitForNavigation();
 $page->waitUntilContainsElement('div[data-name=\"el\"]');
 ```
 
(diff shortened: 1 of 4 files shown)


jhoopmann:
fixxed private const access

enricodias (approved):

GrahamCampbell:
Thanks.

PR #548, opened 2023-09-20
[1.9] Improve stuffs about assertions

peter279k:
# Changed log

- Making the assertion correct argument positions.
- Using the `assertNull` to assert the the expected is same as null.
- Using the `assertSame` to let assertion not be ambiguous.
- Using the `assertCount` to assert the expected countable is same as actual counter.


enricodias:
Thanks!

PR #549, opened 2023-09-20
[1.9] Shorten null coalescing operator

peter279k:
# Changed log

- Shorten the null operator to make it readable for self assignment.


enricodias:
Thanks.

PR #551, opened 2023-10-01
[1.10] Update PHPUnit configuration

osbre:
```
There was 1 PHPUnit test runner deprecation:

1) Your XML configuration validates against a deprecated schema. Migrate your XML configuration using "--migrate-configuration"!
```


GrahamCampbell:
I think this breaks coverage on phpunit 9 though?

osbre:
Tested with PHPUnit 9 - all tests passed but adds one more warning to the beginning of the output:
```
Line 9:
 - Element 'source': This element is not expected.

Test results may not be as expected.
 ```

If I run the tests before this change with PHPUnit 9, it always shows this warning:

```
Line 2:

- Element 'phpunit', attribute 'requireCoverageMetadata': The attribute 'requireCoverageMetadata' is not allowed.

Test results may not be as expected.
```

So, it works but adds one more warning. 

`¯\_(ツ)_/¯`

GrahamCampbell:
Warning is worse than deprecation. ;)

deprecation < warning < error.

GrahamCampbell:
There is more than just a warning there FYI, it is actually broken in loading the source for coverage.

osbre:
Sorry, indeed, I missed the part about coverage 👀

PR #553, opened 2023-10-08
[1.12] Add the ability to screenshot individual elements on the page

osbre:
Note that currently, the position X and Y properties are not being recorded correctly because the type is `int`, while the original X and Y type can be float. This PR fixes it as well.

### DIFF
--- a/src/Clip.php
+++ b/src/Clip.php
@@ -24,17 +24,17 @@ class Clip
      *
      * @param int|float $x
      * @param int|float $y
-     * @param int|float $height
-     * @param int|float $width
-     * @param int|float $scale
+     * @param int       $height
+     * @param int       $width
+     * @param float     $scale
      */
     public function __construct($x, $y, $width, $height, $scale = 1.0)
     {
         $this->x = (float) $x;
         $this->y = (float) $y;
-        $this->height = (float) $height;
-        $this->width = (float) $width;
-        $this->scale = (float) $scale;
+        $this->height = $height;
+        $this->width = $width;
+        $this->scale = $scale;
     }
 
     /**
@@ -54,7 +54,7 @@ class Clip
     }
 
     /**
-     * @return float
+     * @return int
      */
     public function getHeight()
     {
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
This is probably still broken because width and height have to be ints, so you are just getting the top left corner position correct, but not the bottom right?

GrahamCampbell:
Are you able to test the change I pushed up work, please, and confirm that all 4 things are floats (and not just degenerate floats with a .0), passed through to chrome?

osbre:
Tested, here is the data sample:

```php
dd($element->getClip());
```
```
Clip (instance) {
  #x: 8.0
  #y: 79.875
  #height: 21.0
  #width: 784.0
  #scale: 1.0
}
```

Seems correct. (no fake floats)

Thanks!

enricodias:
That's a really simple but extremely useful feature 👍

enricodias (approved):

GrahamCampbell:
The width and height look wrong. Surely those should also be fractional?

osbre:
@GrahamCampbell If you meant that the window `width` and `height` need to be `int` instead of `float` - that's correct according to the code I've tried running. Getting this error if passed float:

```
Invalid parameters - Failed to deserialize params.height - BINDINGS: int32 value expected at position 23
```

osbre:
@GrahamCampbell Sorry for asking, but what do you mean by saying "fractional" in this context?

GrahamCampbell:
Fractional as in floats that are not ending in `.0`

GrahamCampbell:
Awaiting reply from @osbre.

osbre:
@GrahamCampbell Did a test which confirmed that the width/height will be fractional if the element's width/height is explicitly set to a fractional number or is a percentage.

```php
$page->navigate($this->sitePath('fractional.html'))->waitForNavigation();

$element = $page->dom()->querySelector('#fractional');
self::assertEquals(90.5, $element->getClip()->getWidth());
```

```html
<html>
<body>
<p id="fractional" style="color: red; width: 90.5;">
    This element has a fractional width.
</p>
</body>
</html>
```

Should I make it a test and include it in this PR?

PR #573, opened 2023-12-12
[1.10] Add "use HeadlessChromium\Dom\Selector\XPathSelector;"

divinity76:
makes it easier to follow sample code, i had to look it up in the source code myself.


enricodias:
I think this one can go straight to `1.11` if we don't add anything else to `1.10`. No need to push an extra version just for a doc change.

divinity76:
¯\\\_(ツ)_/¯  changed target branch.

PR #576, opened 2023-12-13
[1.10] Fix xpath php-string-to-javascript-string

divinity76:
TL;DR: addslashes() is not the correct way to convert a php-string to a javascript string. json_encode() is.

For example, addslashes will fail on the PHP string "foo".chr(10)."bar" , the old addslashes() code will convert it into "foo
bar"

which is a javascript syntax error.

Previously this code would fail:
```php
$str = "foo".chr(10)."bar";
$xps = new XPathSelector("//span[contains(text(),'" . $str . "')]");
var_dump($xps->expressionCount());
```
it would generate a javascript syntax error:
```
string(134) "document.evaluate("//span[contains(text(),\'foo
bar\')]", document, null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null).snapshot Length"
```
now it generates valid javascript:
```
string(133) "document.evaluate("//span[contains(text(),'foo\nbar')]", document,
null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null).snapshotLength"
```

### DIFF
--- a/src/Dom/Selector/XPathSelector.php
+++ b/src/Dom/Selector/XPathSelector.php
@@ -19,11 +19,11 @@ final class XPathSelector implements Selector
 
     public function expressionCount(): string
     {
-        return 'document.evaluate('. json_encode($this->expression, JSON_THROW_ON_ERROR) .', document, null, XPathResult.ORDERED_NODE_SNAPSHOT_TYPE, null).snapshotLength';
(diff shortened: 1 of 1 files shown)


divinity76:
I wonder if perhaps the original code was written to support PHP older than 5.2.0 (released 2006), where json_encode didn't exist yet

GrahamCampbell:
Ha, I doubt it. The original author probably just misunderstood how to prepare the string.

GrahamCampbell:
My mistake, really. I didn't catch it: https://github.com/chrome-php/chrome/pull/340.

divinity76:
i don't see any reason to postpone merging this

PR #595, opened 2024-01-11
fix guessChromeBinaryPath on Ubuntu Linux + add Chromium

divinity76:
on Ubutu Linux, "command" is a bash builtin, not a standalone binary, and shell_exec() tries to find the "command" binary and fails, thus the original code never worked on Ubuntu Linux (I think the same is true for the entire Debian-derived family of operating systems, but don't quote me on that),  fixing that by doing `bash -c 'command...`  instead.

Also added chromium support, for those who have Chromium but not Chrome installed.

It will look for Chrome first, and if it cannot find Chrome, it will look for Chromium second, and if all fails, it falls back to hardcoded "chrome".

### DIFF
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -40,7 +40,7 @@ class AutoDiscover
             case 'Windows':
                 return self::getFromRegistry() ?? '%ProgramFiles(x86)%\Google\Chrome\Application\chrome.exe';
             default:
-                return \explode("\n", (string) self::shellExec("bash -c 'command -v google-chrome chrome chromium-browser chromium'"), 2)[0] ?: 'chrome';
+                return rtrim(\explode("\n", (string) self::shellExec("bash -c 'command -v google-chrome chrome chromium-browser chromium'"), 2)[0] ?: 'chrome');
         }
     }
 
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -40,7 +40,7 @@ class AutoDiscover
             case 'Windows':
                 return self::getFromRegistry() ?? '%ProgramFiles(x86)%\Google\Chrome\Application\chrome.exe';
             default:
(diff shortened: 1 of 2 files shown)


enricodias:
I use Ubuntu and it works fine here 🤔
```
λ command -v google-chrome
/usr/bin/google-chrome
```

divinity76:
> I use Ubuntu and it works fine here 🤔
> 
> ```
> λ command -v google-chrome
> /usr/bin/google-chrome
> ```

yeah in your shell with your shell builtins, but try
```sh
php -r 'var_dump(shell_exec("command -v google-chrome"));'
```
and then try
```sh
php -r 'var_dump(shell_exec("bash -c '\''command -v google-chrome'\''"));'
```

enricodias:
Same thing in both:

```bash
λ php -r 'var_dump(shell_exec("command -v google-chrome"));'  
string(23) "/usr/bin/google-chrome
"
```

```bash
λ php -r 'var_dump(shell_exec("bash -c '\''command -v google-chrome'\''"));'  
string(23) "/usr/bin/google-chrome
"
```

divinity76:
@enricodias well that confuses me, what do you get by
```
type command;
php -r 'var_dump(shell_exec("which command"));'
```
?

When I do it, I get
```
hans@DESKTOP-EE15SLU:/$ type command; php -r 'var_dump(shell_exec("which command"));'
command is a shell builtin
NULL
```

enricodias:
I got the same;
```bash
command is a shell builtin
NULL
```

It is a builtin, but `shell_exec` can still use it.

divinity76:
@enricodias  wow.. you're entirely correct!

I could've sworn that 
```php -r 'var_dump(shell_exec("command -v google-chrome"));'```
didn't work for me yesterday, and concluded that shell_exec() doesn't have access to shell builtins,
... but it does work now! nevermind 🤷‍♂️ and sorry

PR #597, opened 2024-01-12
[1.11] Autodetect Chromium too

divinity76:
Previously only google-chrome was detected, that's a nuisance for people who only have Chromium installed.


GrahamCampbell:
Thanks.

PR #599, opened 2024-01-17
Adding support for http endpoint in addition to the websocket

vrobert78:
In order to use Chrome Headless you need to know the `webSocketDebuggerUrl`.
This address can be retrieved by requesting `/json/version` as mentioned at https://chromedevtools.github.io/devtools-protocol/.

With this PR the goal is to support both the websocket address and the HTTP endpoint.

### DIFF
--- a/src/Communication/Connection.php
+++ b/src/Communication/Connection.php
@@ -107,19 +107,19 @@ class Connection extends EventEmitter implements LoggerAwareInterface
         // create socket client
         if (\is_string($socketClient)) {
 
-            if (\in_array(\parse_url($socketClient, PHP_URL_SCHEME), ['http', 'https'])) {
+            if (\in_array(\parse_url($socketClient, \PHP_URL_SCHEME), ['http', 'https'])) {
 
                 $configURL = $socketClient.'/json/version';
 
                 if (!\function_exists('curl_init')) {
-                    throw new \Exception("curl is not available");
+                    throw new \Exception('curl is not available');
                 }
 
-                $curl = curl_init($configURL);
-                curl_setopt($curl, CURLOPT_URL, $configURL);
-                curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
-                $resp = curl_exec($curl);
-                curl_close($curl);
+                $curl = \curl_init($configURL);
+                \curl_setopt($curl, CURLOPT_URL, $configURL);
+                \curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
+                $resp = \curl_exec($curl);
+                \curl_close($curl);
 
                 if (false === $resp) {
                     throw new \Exception("Unable to request $configURL");
--- a/src/Communication/Connection.php
+++ b/src/Communication/Connection.php
@@ -107,33 +107,23 @@ class Connection extends EventEmitter implements LoggerAwareInterface
         // create socket client
         if (\is_string($socketClient)) {
 
-            $parsedUrl = parse_url($socketClient);
-            if (isset($parsedUrl['scheme']) && ($parsedUrl['scheme']==='http' or $parsedUrl['scheme']==='https'))
+            if (\in_array(\parse_url($socketClient, PHP_URL_SCHEME), ['http', 'https']))
             {
 
                 $configURL = $socketClient.'/json/version';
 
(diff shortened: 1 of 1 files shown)


src/Communication/Connection.php
```diff
@@ -106,6 +106,41 @@ public function __construct($socketClient, LoggerInterface $logger = null, int $
 
         // create socket client
         if (\is_string($socketClient)) {
+
+            $parsedUrl = parse_url($socketClient);
+            if (isset($parsedUrl['scheme']) && ($parsedUrl['scheme']==='http' or $parsedUrl['scheme']==='https'))
```
enricodias: The expression could be simpler:
```suggestion
            if (\in_array(\parse_url($socketClient, PHP_URL_SCHEME), ['http', 'https'])
```

src/Communication/Connection.php
```diff
@@ -106,6 +106,41 @@ public function __construct($socketClient, LoggerInterface $logger = null, int $
...
+                curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
+                $resp = curl_exec($curl);
+
+                if ($resp === false) {
+                    $err = curl_error($curl);
+                    curl_close($curl);
+
+                    throw new \Exception("Unable to request $configURL");
+                }
+                else {
```
enricodias: This else isn't doing anything. The throw in the previous if will stop the execution on this method.

src/Communication/Connection.php
```diff
@@ -106,6 +106,41 @@ public function __construct($socketClient, LoggerInterface $logger = null, int $
...
+
+                $configURL = $socketClient.'/json/version';
+
+                $curl = curl_init($configURL);
+                curl_setopt($curl, CURLOPT_URL, $configURL);
+                curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
+                $resp = curl_exec($curl);
+
+                if ($resp === false) {
+                    $err = curl_error($curl);
```
enricodias: Why get the error if we are not using it for anything?

enricodias:
Thanks for the contribution. I left some comments in the code.
It also wouldn't hurt checking if curl is available before attempting to use it since this lib doesn't use it anywhere else and there are some styleci issues to be resolved.

vrobert78:
Hi @enricodias, thanks a lot for your feedback. I just committed a few changes from your recommendations.

src/Communication/Connection.php
```diff
@@ -106,6 +106,30 @@ public function __construct($socketClient, LoggerInterface $logger = null, int $
 
         // create socket client
         if (\is_string($socketClient)) {
+
+            if (\in_array(\parse_url($socketClient, PHP_URL_SCHEME), ['http', 'https'])) {
+
+                $configURL = $socketClient.'/json/version';
+
+                $resp = \file_get_contents($configURL);
```
GrahamCampbell: We can't call this without doing any error handling (yes, I really do mean "error" in the PHP 5 sense of the word).

enricodias:
Some environments will prevent file functions from accessing external urls. I could just check if curl is available and use it.

vrobert78:
> Some environments will prevent file functions from accessing external urls. I could just check if curl is available and use it.

Hey @enricodias , curl is back and I've added a check for it.

enricodias (approved):
Looks good to me now.

src/Communication/Connection.php
divinity76: please always use strict mode, unless you have a very good reason to not do it - it's less important in modern php, but in <=7.4: https://3v4l.org/Yc3UU

src/Communication/Connection.php
```diff
@@ -106,6 +106,35 @@ public function __construct($socketClient, LoggerInterface $logger = null, int $
 
         // create socket client
         if (\is_string($socketClient)) {
+            if (\in_array(\parse_url($socketClient, \PHP_URL_SCHEME), ['http', 'https'])) {
```
osbre, line 109: Wouldn't it be better to extract all this code into a `private/protected` method? something like 
```php
private function retrieveWsDebuggerUrl(string $endpoint): string
```

GrahamCampbell:
I'm not going to go ahead with this.

Issue #613, opened 2024-03-19
command not (always?) available in shell_exec


divinity76:
testing with php8.3 on ubuntu24.04-beta, `command` is not available in shell_exec:
```sh
hans@DESKTOP-EE15SLU:~/projects/headless-chomium-recorder$ command -v google-chrome chromium-browser chrome chromium
/usr/bin/chromium-browser
/snap/bin/chromium
hans@DESKTOP-EE15SLU:~/projects/headless-chomium-recorder$ php -r 'var_dump(shell_exec("command -v google-chrome chromium-browser chrome chromium"));'
Command line code:1:
NULL
hans@DESKTOP-EE15SLU:~/projects/headless-chomium-recorder$ php -r 'var_dump(shell_exec("bash -c '\''command -v google-chrome chromium-browser chr
ome chromium'\''"));'
Command line code:1:
string(45) "/usr/bin/chromium-browser
/snap/bin/chromium
"
hans@DESKTOP-EE15SLU:~/projects/headless-chomium-recorder$ php -r 'var_dump(shell_exec("which google-chrome chromium-browser chrome chromium"));'
Command line code:1:
string(45) "/usr/bin/chromium-browser
/snap/bin/chromium
"
hans@DESKTOP-EE15SLU:~/projects/headless-chomium-recorder$ 
```

PHP8.3 migrated some execution commands from fork+exec to posix_spawn, this might be another side-effect? 

either way, seems our alternatives are to migrate to `type (...)` or `bash -c 'command -v (...)'` or implement our own `getenv("PATH")`-scanner (or just find one in composer) 🤔 

this breaks chrome/chromium path autodetect in BrowserFactory

divinity76:
proposed fix: https://github.com/chrome-php/chrome/pull/614 
(manual PATH scanner)

PR #614, opened 2024-03-19
manual PATH scanner instead of command

divinity76:
command is not always available in shell_exec() , observed on Ubuntu-24.04-beta + PHP8.3
see https://github.com/chrome-php/chrome/issues/613 for details

### DIFF
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -59,7 +59,6 @@ class AutoDiscover
 
                 return 'chrome'; // ... very unlikely to actually work, but this retains the original behavior..
                 // this would make more sense: throw new \RuntimeException('Could not find chrome binary');
-            }
         }
     }
 
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -53,12 +53,10 @@ class AutoDiscover
                             return $file;
                         }
                     }
+
                     return 'chrome'; // ... very unlikely to actually work, but this retains the original behavior..
                     throw new \RuntimeException('Could not find chrome binary'); // this makes more sense tbh
                 }
-
-                return 'chrome'; // ... very unlikely to actually work, but this retains the original behavior..
-                // this would make more sense: throw new \RuntimeException('Could not find chrome binary');
         }
     }
 
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -55,7 +55,7 @@ class AutoDiscover
                     }
 
                     return 'chrome'; // ... very unlikely to actually work, but this retains the original behavior..
-                    throw new \RuntimeException('Could not find chrome binary'); // this makes more sense tbh
+                    // throw new \RuntimeException('Could not find chrome binary'); // this makes more sense tbh
                 }
         }
     }
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -53,10 +53,9 @@ class AutoDiscover
                             return $file;
                         }
                     }
-
-                    return 'chrome'; // ... very unlikely to actually work, but this retains the original behavior..
-                    // throw new \RuntimeException('Could not find chrome binary'); // this makes more sense tbh
                 }
+                return 'chrome'; // ... very unlikely to actually work, but this retains the original behavior..
+                // throw new \RuntimeException('Could not find chrome binary'); // this makes more sense tbh
         }
     }
 
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -39,23 +39,26 @@ class AutoDiscover
                 return '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome';
             case 'Windows':
                 return self::getFromRegistry() ?? '%ProgramFiles(x86)%\Google\Chrome\Application\chrome.exe';
-            default: {
-                    $valid_names = array(
-                        'google-chrome',
-                        'chromium-browser',
-                        'chrome',
-                        'chromium',
-                    );
-                    foreach (\explode(\PATH_SEPARATOR, \getenv('PATH')) as $dir) {
-                        foreach ($valid_names as $name) {
-                            $file = $dir . \DIRECTORY_SEPARATOR . $name;
-                            if (\is_file($file) && \is_executable($file)) {
-                                return $file;
-                            }
+            default:
+                $valid_names = [
+                    'google-chrome',
+                    'chromium-browser',
+                    'chrome',
+                    'chromium',
+                ];
+                foreach (\explode(\PATH_SEPARATOR, \getenv('PATH')) as $dir) {
+                    foreach ($valid_names as $name) {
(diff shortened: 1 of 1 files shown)


src/AutoDiscover.php
```diff
@@ -40,7 +40,23 @@ public function guessChromeBinaryPath(): string
...
+                ];
+                foreach (\explode(\PATH_SEPARATOR, \getenv('PATH')) as $dir) {
+                    foreach ($valid_names as $name) {
+                        $file = $dir.\DIRECTORY_SEPARATOR.$name;
+                        if (\is_file($file) && \is_executable($file)) {
+                            return $file;
+                        }
+                    }
+
+                    return 'chrome'; // ... very unlikely to actually work, but this retains the original behavior..
```
enricodias: That will return `chrome` after checking only the first folder. Didn't work here since my chrome is not in my first path folder. The fallback must be outside both loops.

enricodias:
My `PATH` has 21 folders, that would result in 84 checks (21 x 4) in the worst case. I know that `command -v` probably does the same thing under the hood, but it feels so wrong to scan folders manually like this 😅

divinity76:
> My `PATH` has 21 folders, that would result in 84 checks (21 x 4) in the worst case. I know that `command -v` probably does the same thing under the hood, but it feels so wrong to scan folders manually like this 😅

yeah i know what you mean😅 i have 25 here (WSL Ubuntu 24.04-beta)

Benchmarking it on a AMD Ryzen 9 7950x (high-end 2022 system) Ubuntu 24.04-beta WSL:
```
$ sudo bash -c 'sync; echo 3 > /proc/sys/vm/drop_caches';
$ time bash -c ' command -v google-chrome chromium-browser chrome chromium'
/usr/bin/chromium-browser
/snap/bin/chromium

real    0m0.024s
user    0m0.003s
sys     0m0.000s
$ time bash -c ' command -v google-chrome chromium-browser chrome chromium'
/usr/bin/chromium-browser
/snap/bin/chromium

real    0m0.022s
user    0m0.003s
sys     0m0.000s
$ sudo bash -c 'sync; echo 3 > /proc/sys/vm/drop_caches';
$ time php -r '                                          
$a=explode(PATH_SEPARATOR, getenv("PATH"));
var_dump(count($a));
$names=["google-chrome","chromium-browser","chrome","chromium"];
foreach($a as $dir){
  foreach($names as $name){
    $file=$dir.DIRECTORY_SEPARATOR.$name;
    if(file_exists($file) && is_executable($file)){
      echo $file."\n";
    }
  }
}
';
Command line code:3:
int(25)
/usr/bin/chromium-browser
/bin/chromium-browser
/snap/bin/chromium

real    0m0.097s
user    0m0.015s
sys     0m0.007s
$ time php -r '
$a=explode(PATH_SEPARATOR, getenv("PATH"));
var_dump(count($a));
$names=["google-chrome","chromium-browser","chrome","chromium"];
foreach($a as $dir){
  foreach($names as $name){
    $file=$dir.DIRECTORY_SEPARATOR.$name;
    if(file_exists($file) && is_executable($file)){
      echo $file."\n";
    }
  }
}
';
Command line code:3:
int(25)
/usr/bin/chromium-browser
/bin/chromium-browser
/snap/bin/chromium

real    0m0.041s
user    0m0.015s
sys     0m0.000s
```
(php does more work than command here because php does not bail, it simulates absolutely worst case scenario, command does not)

doing the same on a Intel i7-8565U, a 2018 **1**5watt low-power laptop cpu running on battery in power-saving mode, on WSL Ubuntu 24.04-beta:
```sh
hans@DESKTOP-2LHJILI:~$ sudo bash -c 'sync; echo 3 > /proc/sys/vm/drop_caches';
hans@DESKTOP-2LHJILI:~$ time bash -c ' command -v google-chrome chromium-browser chrome chromium'
/usr/bin/chromium-browser
/snap/bin/chromium

real    0m0.820s
user    0m0.001s
sys     0m0.050s
hans@DESKTOP-2LHJILI:~$ time bash -c ' command -v google-chrome chromium-browser chrome chromium'
/usr/bin/chromium-browser
/snap/bin/chromium

real    0m0.476s
user    0m0.001s
sys     0m0.027s
hans@DESKTOP-2LHJILI:~$ sudo bash -c 'sync; echo 3 > /proc/sys/vm/drop_caches';
hans@DESKTOP-2LHJILI:~$ time php -r '
$a=explode(PATH_SEPARATOR, getenv("PATH"));
var_dump(count($a));
$names=["google-chrome","chromium-browser","chrome","chromium"];
foreach($a as $dir){
  foreach($names as $name){
    $file=$dir.DIRECTORY_SEPARATOR.$name;
    if(file_exists($file) && is_executable($file)){
      echo $file."\n";
    }
  }
}
';
int(32)
/usr/bin/chromium-browser
/bin/chromium-browser
/snap/bin/chromium

real    0m1.168s
user    0m0.062s
sys     0m0.076s
hans@DESKTOP-2LHJILI:~$ time php -r '
$a=explode(PATH_SEPARATOR, getenv("PATH"));
var_dump(count($a));
$names=["google-chrome","chromium-browser","chrome","chromium"];
foreach($a as $dir){
  foreach($names as $name){
    $file=$dir.DIRECTORY_SEPARATOR.$name;
    if(file_exists($file) && is_executable($file)){
      echo $file."\n";
    }
  }
}
';
int(32)
/usr/bin/chromium-browser
/bin/chromium-browser
/snap/bin/chromium

real    0m0.617s
user    0m0.030s
sys     0m0.030s
```
🤔 i think servers will do just fine. laptop users should specify chrome path to avoid the delay

src/AutoDiscover.php
divinity76: 😅 i messed up while trying to appease StyleCI via github web interface

divinity76:
could be cached, but then we'd need to consider cache invalidation and [cache poisoning](https://en.wikipedia.org/wiki/Cache_poisoning) 🤔

enricodias:
The performance will always be ok since `command` do the same thing, it just _feels_ wrong 😅. The code itself seems to work. If I have time in the next few days I'll see if there is any other better way of doing this.

Php already has the stat cache. It won't do anything in this case but it's still not an issue.

GrahamCampbell:
I'm not really sure we should be making auto-discover that complicated. The point is to detect the most common places, only. I feel like people should be passing in the binary path explicitly, otherwise, or implementing this kind of logic in their own app.

divinity76:
@GrahamCampbell so you're basically proposing
```
$commonLocations = array(
"/usr/bin/chromium-browser",
"/bin/chromium-browser",
"/snap/bin/chromium",
(...)
);
ForEach($commonLocations as (...)
``` 
?

enricodias:
I couldn't think of any better way to implement this and I recreating the `command` behaviour seems out of the scope of this project, but if you manage to convince Graham I'm ok with it.

GrahamCampbell:
I'm not going to go ahead with this.

Issue #469, opened 2023-01-14
Error: Could not find node with given id


santossoares:
I got an issue with this library. I think its a bug.

When i use the  querySelectorAll function, and iterate over its elements, the code works fine!

While on the loop iterating with the elements, and using another selector on the $page instance, then it looses the "reference to the elements", it throws the following error: Could not find node with given id.

Can anyone solve this?

Thanks!

Here is the sample code:

```php
use HeadlessChromium\BrowserFactory;

$browserFactory = new BrowserFactory();

$browser = $browserFactory->createBrowser([
    'headless' => false
]);

$page = $browser->createPage();
$page->navigate('http://google.com')->waitForNavigation();

$elements = $page->dom()->querySelectorAll('.pHiOh');

foreach ($elements as $e) {
    echo($e->getText()).PHP_EOL;
    
    // This new selector causes error on the previous line!
    // Just remove this line and the code works!
    $buttons = $page->dom()->querySelectorAll('button');

    // ERROR: Uncaught HeadlessChromium\Exception\DomException: -32000 - Could not find node with given id
}

sleep(3);
$browser->close();
```

aravael:
Hi @santossoares , your remark is fair enough. I noticed such behavior a while ago and I think it's because `$page->dom()` creates a new Dom object every time it is invoked. Definitely need clarification from @enricodias .
In the meanwhile a workaround is this:
```
use HeadlessChromium\BrowserFactory;

$browserFactory = new BrowserFactory();

$browser = $browserFactory->createBrowser([
    'headless' => false
]);

$page = $browser->createPage();
$page->navigate('https://google.com')->waitForNavigation();

$elements = $page->dom()->querySelectorAll('.pHiOh');

for ($i = 0; $i < count($elements); $i++) {
    $elements = $page->dom()->querySelectorAll('.pHiOh');

    echo $elements[$i]->getText();

    $buttons = $page->dom()->querySelectorAll('button');
}
```

santossoares:
Thanks for the reply! I got some workaround too!
I just call the $page->dom() only once, with for example body element, and store it in a variable.
Then if i want to search for another selector, i use that variable that contains the body element and search using the queryselector from the variable.

$body = $page->dom('body');
$body->querySelectorAll('.someclass');

This way, the dom is never replaced like when you call the $page->dom() again.

enricodias:
[...]

PR #620, opened 2024-04-23
Fix could not find node with given id

Seriyyy95:
It is a fix for issue [469](https://github.com/chrome-php/chrome/issues/469)

Previously, a new `Dom` instance was created for each call of `$page->dom()`. In this pull request, I have stored `$dom` as a variable in the `Page` class and added an auto-update of the root node id using the `DOM.documentUpdated` event. 

I use `waitForData()` from upcoming `chrome-php/wrench` 1.7 to process all events before each request without blocking execution if there is no data to process. When the page is refreshed in Chrome, chrome-php will be notified about it before making any requests. Somehow, Chrome fires two `DOM.documentUpdated` events during one reload, so I decided to not fetch the node ID in the event handler but just mark the node as stale and request a new node ID before making a request if the old one is stale.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
+.idea
+
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,4 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
-.idea
 
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
-
--- a/composer.json
+++ b/composer.json
@@ -46,8 +46,6 @@
         },
         "preferred-install": "dist"
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
     "extra": {
         "bamarni-bin": {
             "bin-links": true,
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     ],
     "require": {
         "php": "^7.4.15 || ^8.0.2",
-        "chrome-php/wrench": "^1.6",
+        "chrome-php/wrench": "^1.7",
         "evenement/evenement": "^3.0.1",
         "monolog/monolog": "^1.27.1 || ^2.8 || ^3.2",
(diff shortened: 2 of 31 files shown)


.gitignore
```diff
@@ -3,3 +3,5 @@ composer.lock
 phpstan.neon
 phpunit.xml
 vendor
+.idea
```
GrahamCampbell: please revert this

composer.json
```diff
@@ -46,6 +46,8 @@
         },
         "preferred-install": "dist"
     },
+    "minimum-stability": "dev",
```
GrahamCampbell: please revert this

src/Communication/Socket/MockSocket.php
```diff
@@ -124,4 +124,11 @@ public function disconnect($reason = 1000)
 
         return true;
     }
+
+    public function waitForData(float $maxSeconds): bool
+    {
+        // TODO: Implement wait if needed.
```
GrahamCampbell: probably we don't need this

src/Communication/Socket/MockSocket.php
```diff
@@ -124,4 +124,11 @@ public function disconnect($reason = 1000)
...
+
+        return !empty($this->receivedData);
```
GrahamCampbell: please don't use the empty function

src/Communication/Socket/SocketInterface.php
```diff
@@ -52,4 +52,6 @@ public function isConnected();
      * @return bool
      */
     public function disconnect($reason = 1000);
+
+    public function waitForData(float $maxSeconds): bool;
```
GrahamCampbell: adding this is a breaking change. can you make another interface that just has waitForData, implement it, and have callers use an instanceof check please.

GrahamCampbell (changes requested):

composer.json
Seriyyy95: This line is needed to be able to install `chrome-php/wrench` 1.7. I removed it in the last commit, and now tests won't pass. I guess this will be fixed after the `wrench` release.

src/Communication/Socket/MockSocket.php
Seriyyy95: I have deleted implementation of `waitForData()` from `MockSocket` because it is not used and not needed to implement the interface.

PR #624, opened 2024-05-14
[1.13] `getBase64` returns string

divinity76:
... reading the code, it suggest the return can either be string or null, but with the isSuccessful() check, i think it can only ever return string?

not 100% sure if it should be `string` or `?string`  but we can narrow it down from `mixed`.

### DIFF
--- a/src/PageUtils/AbstractBinaryInput.php
+++ b/src/PageUtils/AbstractBinaryInput.php
@@ -41,9 +41,9 @@ abstract class AbstractBinaryInput
     /**
      * Get base64 representation of the file.
      *
-     * @return mixed
+     * @return string
      */
-    public function getBase64(int $timeout = null)
+    public function getBase64(int $timeout = null): string
     {
         $response = $this->responseReader->waitForResponse($timeout);
 


enricodias:
It should only be string but technically changing types to a more restrict one is still a breaking change :/
There are lot's of other cases like that in this lib. Changing them alone is not enough reason for a new major version, but we should fix them all in the first opportunity we have.

divinity76:
@enricodias got an idea! what if we make a `develop` branch where we queue up all the changes that should be part of the next release, and branch out from develop for making actual releases?

enricodias:
We already use the main branch as the current developing branch and make releases from it. Having another branch for a 2.x version would require us to keep syncing new features and fixes from 1.x to 2.x until the new version actually gets released. People also don't usually update major versions that often because they have breaking changes. We would end up having to support 2 versions for some time. When 1.0 was released, it took several months for the majority of users to migrate from 0.x to it. I think we need a better reason to justify a whole new major version, like a rework of some core features.

GrahamCampbell:
I'm going to leave this for now.

PR #504, opened 2023-03-31
[1.11] Add support for the new Chrome headless mode

sanderdlm:
Fixes https://github.com/chrome-php/chrome/issues/498

Reference: https://developer.chrome.com/articles/new-headless/

Imlemented just like Puppeteer:
```
To opt in to the new Headless mode in Puppeteer:

import puppeteer from 'puppeteer';

const browser = await puppeteer.launch({
  headless: 'new',
  // `headless: true` (default) enables old Headless;
  // `headless: 'new'` enables new Headless;
  // `headless: false` enables “headful” mode.
});

const page = await browser.newPage();
await page.goto('https://developer.chrome.com/');

// …

await browser.close();
```

Passing true or not setting the flag at all still enables the old mode. False still disables the headless mode. Passing "new" enables the new headless mode.

From the blog post:

> Currently, passing the --headless command-line flag without an explicit value still activates the old Headless mode — but we plan to change this default to new Headless over time.

This code will work during the transition period and allow users to test their applications against the new set-up

> We plan to completely remove the old Headless from the Chrome binary and stop supporting this mode in Puppeteer later this year. As part of this removal, we’ll make the old Headless available as a separate standalone binary for those users who can’t upgrade yet.

When the old headless mode is removed, the default in this library might have to be switched to "new" (or Chrome will simply map --headless to --headless=new internally, we don't know yet)

PR #628, opened 2024-06-07
Add new headless to lanuch options

mrmmg:
Recently, I used this package to scrape a website that requires users to log in.

My target website uses cookies, IndexedDB, and SessionStorage, and sometimes, even with a logged-in browser with user-data-directory, I still see the login form. In the DevTools of headless chrome , there are many errors indicating that Chrome cannot access cookie storage or IndexedDB. After this, I tried crawling with Node.js and the Puppeteer library, but the issue remained. I then started browsing the Puppeteer issues and found some similar to mine:

puppeteer/puppeteer#12498

puppeteer/puppeteer#1316

puppeteer/puppeteer#1268

puppeteer/puppeteer#1270

puppeteer/puppeteer#5612

[Chrome Developers Docs About headless=new](https://developer.chrome.com/docs/chromium/new-headless)

In the first link, a user said that by setting the Puppeteer launch option to headless: 'new', the issue was resolved. I did the same and set the headless mode to 'new', which solved my problems. However, when I wanted to try the new headless mode with the chrome-php/chrome package, I realized that the library does not support this feature. So, I created this pull request and made the necessary changes.

About Tests
I wrote only two tests, as I believe that is enough to test the new code. However, if you think there are any missing points, please let me know, and I will add more tests.

About Documentation
I am waiting for this pull request to be merged into the main code, after which I will add proper documentation for the new headless mode feature.

### DIFF
--- a/src/Browser/BrowserProcess.php
+++ b/src/Browser/BrowserProcess.php
@@ -324,12 +324,14 @@ class BrowserProcess implements LoggerAwareInterface
         }
 
         // enable headless mode
-        if (!\array_key_exists('headless', $options) || $options['headless']) {
-            $args[] = '--headless';
-            $args[] = '--disable-gpu';
-            $args[] = '--font-render-hinting=none';
-            $args[] = '--hide-scrollbars';
-            $args[] = '--mute-audio';
+        if (!array_key_exists('headless', $options) || $options['headless']) {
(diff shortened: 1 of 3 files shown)


mrmmg:
I noticed that this feature is already done by a user in PR #504, So I decide to close this PR.

PR #655, opened 2024-10-25
[1.15] Fix precision loss (int vs float mixup)

divinity76:
The origin X/Y comes from Page.getLayoutMetrics.cssContentSize ( https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-getLayoutMetrics )  which returns float values like -0.666748046875 , not integers, but the library erroneously treated it like integers, that can cause Mouse()->find(...)->click() to miss-click on very small targets, and with custom error handlers it can also cause
```
["trace":"Exception":private]=>
    array(9) {
      [0]=>
      array(4) {
        ["file"]=>
        string(52) "C:\temp\vendor\chrome-php\chrome\src\Input\Mouse.php"
        ["line"]=>
        int(338)
        ["function"]=>
        string(9) "{closure}"
        ["args"]=>
        array(4) {
          [0]=>
          int(8192)
          [1]=>
          string(69) "Implicit conversion from float -0.666748046875 to int loses precision"
          [2]=>
          string(52) "C:\temp\vendor\chrome-php\chrome\src\Input\Mouse.php"
          [3]=>
          int(338)
        }
      }
```
(and that is exactly what happened to me and how I noticed the problem)

### DIFF
--- a/src/Input/Mouse.php
+++ b/src/Input/Mouse.php
@@ -31,8 +31,8 @@ class Mouse
      */
     protected $page;
 
-    protected $x = 0;
-    protected $y = 0;
+    protected $x = 0.0;
+    protected $y = 0.0;
 
     protected $button = self::BUTTON_NONE;
 
@@ -54,7 +54,7 @@ class Mouse
      *
      * @return $this
      */
-    public function move(int $x, int $y, array $options = null)
+    public function move(float $x, float $y, array $options = null)
     {
         $this->page->assertNotClosed();
 
@@ -329,13 +329,13 @@ class Mouse
     /**
      * Get the maximum distance to scroll a page.
      *
-     * @param int $distance Distance to scroll, positive or negative
-     * @param int $current  Current position
-     * @param int $maximum  Maximum possible distance
+     * @param float $distance Distance to scroll, positive or negative
+     * @param float $current  Current position
+     * @param float $maximum  Maximum possible distance
      *
-     * @return int allowed distance to scroll
+     * @return float allowed distance to scroll
      */
-    private function getMaximumDistance(int $distance, int $current, int $maximum): int
+    private function getMaximumDistance(float $distance, float $current, float $maximum): float
     {
         $result = $current + $distance;
 
--- a/src/Input/Mouse.php
+++ b/src/Input/Mouse.php
@@ -45,8 +45,8 @@ class Mouse
     }
 
     /**
-     * @param int        $x
-     * @param int        $y
+     * @param float        $x
+     * @param float        $y
      * @param array|null $options
      *
      * @throws \HeadlessChromium\Exception\CommunicationException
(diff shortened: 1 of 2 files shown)


divinity76:
is it possible that float scroll position is a recent change in chrome itself? seems much of the code is written with the expectation that scroll position is int, not float 🤔 but window.scrollY is float (and very floaty like X.5), not int

divinity76:
fixing this actually requires breaking the public api a little bit: $page->mouse()->getPosition() 
used to return `array(int, int);`  but when fixing this it returns `array(float, float)` 🤔

enricodias:
This would change the public api, although changing to float shouldn't be a breaking change and old implementations using integers should still work 🤔

But if those tests shouldn't fail.

divinity76:
@enricodias fixed the testcases. (some were expecting int but getting float. no failures beyond that)

enricodias (approved):

src/Input/Mouse.php
```diff
@@ -45,16 +45,16 @@ public function __construct(Page $page)
...
+     * @param float      $y
      * @param array|null $options
      *
      * @throws \HeadlessChromium\Exception\CommunicationException
      * @throws \HeadlessChromium\Exception\NoResponseAvailable
      *
      * @return $this
      */
-    public function move(int $x, int $y, ?array $options = null)
+    public function move(float $x, float $y, ?array $options = null)
```
GrahamCampbell, line 57: This is a breaking change, when the caller has strict types enabled.

GrahamCampbell (changes requested):

src/Input/Mouse.php
divinity76: commented on wrong line, but yes. the actual api break is on line 34, 35 and 338. 
strict_types have an exception for implicit int->float, so line 57 is not an api break: https://3v4l.org/gS1b7
thus function move can be upgraded to support float without breaking api,
but function getPosition() and getMaximumDistance() cannot. 

Would you prefer to only fix the
>Implicit conversion from float -0.666748046875 to int loses precision

issue, at least for the remainder of v1?

enricodias: Why is `getMaximumDistance` a breaking change if it's a private method? And `getPosition` returns an array with no defined types, not even in the docblock.

divinity76: @enricodias 
> Why is `getMaximumDistance` a breaking change if it's a private method?

sorry you're right, getMaximumDistance is not a breaking change, my bad.

>And getPosition returns an array with no defined types, not even in the docblock.

that's kinda a api-break because code that expect `$page->mouse()->getPosition()['y']` to be int may be disappointed. (it broke tests/MouseApiTest.php line 85 test case)

IMO it's a minor API break tho. getPosition()=>["x"=>int,"y"=>int] becomes ["x"=>float,"y"=>float].

PR #689, opened 2025-04-28
[1.15] Fix Linux Chrome binary detection on POSIX shells

xushulang:
Some Linux distributions use dash as the default sh, but in "dash", "command -v" only supports a single argument.

### DIFF
--- a/src/AutoDiscover.php
+++ b/src/AutoDiscover.php
@@ -40,7 +40,7 @@ class AutoDiscover
             case 'Windows':
                 return self::getFromRegistry() ?? '%ProgramFiles(x86)%\Google\Chrome\Application\chrome.exe';
             default:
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks for reporting this. Are you able to replicate this not working, with a test?

GrahamCampbell:
Are there any downsides to making this change?

xushulang:
> Thanks for reporting this. Are you able to replicate this not working, with a test?

tested already

docker, php:8.4-fpm
image

xushulang:
> Are there any downsides to making this change?

no

divinity76:
#614 would've portably fixed it fwiw, scanning manually instead of using the shell

aprat84:
Hi!

How was this not implemented? It's still an issue in Alpine Linux, which uses Busybox Ash shell.

Should we propose a new PR? @GrahamCampbell

aprat84:
@GrahamCampbell Sorry to ping you again, but could you consider merging?

Issue #696, opened 2025-06-08
Custom referrer


waahhhh:
We have the ability to solve issues https://github.com/chrome-php/chrome/issues/530 and https://github.com/chrome-php/chrome/issues/596.
Chrome allows you to define a custom referrer.
See https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-navigate.

An example using a custom referrer modifying `\HeadlessChromium\PageUtils\PageNavigation::__construct()`:

```php
$this->navigateResponseReader = $page->getSession()->sendMessage(
    new Message('Page.navigate', ['url' => $url, 'referrer' => $referrer])
);
```

Note that the parameter `referrer` should not be set by default.
A value like `NULL` or an empty String causes an error.

It would probably be helpful to use a method that allows you to define the parameters of `Page.navigate` on your own.
This would make it more flexible, which is necessary if you want to customize the policy.
See https://developer.chrome.com/blog/referrer-policy-new-chrome-default?hl=en for more information.

waahhhh:
The problem has not yet been solved.

PR #701, opened 2025-07-13
custom referrer

waahhhh:
Here is a suggestion for solving the Issue https://github.com/chrome-php/chrome/issues/696.


waahhhh:
the feature is still missing.

GrahamCampbell:
I've implemented this in an alternative way at https://github.com/chrome-php/chrome/pull/740.

Issue #666, opened 2024-11-20
Some website return some invalid UTF16 ??


momala454:
Hi
I'm receiving the following error
Uncaught exception 'HeadlessChromium\Exception\CommunicationException\CannotReadResponse' with message Response from chrome remote interface is not a valid json response. JSON error: 10 

JSON error 10 means
```
JSON_ERROR_UTF16 | Malformed UTF-16 characters, possibly incorrectly encoded
```

The exception is thrown on 
vendor/chrome-php/chrome/src/Communication/Connection.php
Method dispatchMessage

The data is from a message `Fetch.requestPaused`, when it's trying to send a tracking data from `https://www.smithsfoodanddrug.com`.
The post data contains unreadable data like this, which prevent the library from doing json_decode over it
```
"postData":"\t\u0000\u0000\u0000\u0011i\u000e\u0006\n&J+\u0.....
```

divinity76:
Can you provide a small reproduce script? What OS are you on? what PHP version are you running? 

I failed to reproduce the issue with:
```php
<?php
declare(strict_types=1);
require_once('vendor/autoload.php');
$chromiumPath = '/usr/bin/chromium-browser';
$factory = new \HeadlessChromium\BrowserFactory($chromiumPath);
$browser = $factory->createBrowser();
$page = $browser->createPage();
$page->navigate('https://www.smithsfoodanddrug.com')->waitForNavigation();
var_dump($page->getHtml());
```
it produced
```
$ php repro.php 
string(244597) "<html dir="ltr" lang="en"><head>
  <meta charset="utf-8">
  <meta name="color-scheme" content="light dark">
  <meta name="theme-color" content="#fff">
(...)
```

momala454:
linux, php 8.1,
I think it's only on their login page `https://www.smithsfoodanddrug.com/signin?redirectUrl=/`, and it happens randomly I think

divinity76:
@momala454 still unable to reproduce, i tried
```php
for($i=0;$i<100;++$i) {
    $page->navigate('https://www.smithsfoodanddrug.com/signin?redirectUrl=/')->waitForNavigation();
    var_dump($page->getHtml());
}
```
and I think it ran like 30 times before I got IP banned, 

but even with~30 iterations, I did not get the invalid UTF16 error you speak of.

Can you please write a script to reproduce the issue?

momala454:
the issue occurs with `Fetch.requestPaused`
so you need to enable 
```php
$page
[...]

PR #703, opened 2025-07-29
add json error message

divinity76:
while looking into https://github.com/chrome-php/chrome/issues/666 i got
> PHP Fatal error:  Uncaught HeadlessChromium\Exception\CommunicationException\CannotReadResponse: Response from chrome remote interface is not a valid json response. JSON error: 10 in /home/hans/projects/misc/vendor/chrome-php/chrome/src/Communication/Connection.php:382

But the message string may contain more valuable data, such as

>Single unpaired UTF-16 surrogate in unicode escape

There is not a specific json error id for "unpaired UTF-16 surrogate",
thus that information was lost prior to this patch.

### DIFF
--- a/src/Communication/Connection.php
+++ b/src/Communication/Connection.php
@@ -379,7 +379,7 @@ class Connection extends EventEmitter implements LoggerAwareInterface
         $jsonErrorId = \json_last_error();
         if (\JSON_ERROR_NONE !== $jsonErrorId) {
             if ($this->isStrict()) {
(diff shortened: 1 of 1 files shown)


enricodias (approved):

GrahamCampbell:
I'd prefer this approach: https://github.com/chrome-php/chrome/pull/705.

divinity76:
> I'd prefer this approach: #705.

LGTM

PR #713, opened 2025-10-03
Test against php 8.5

sergiy-petrov:
* Test against php 8.5

* Update php version for phpstan to 8.4


GrahamCampbell:
Thanks, but PHPStan is intentionally run against PHP 7.4.

PR #720, opened 2025-12-27
[1.15] Add support for PHP 8.5 and Symfony 8

GrahamCampbell:


composer.json
```diff
@@ -17,18 +17,18 @@
     ],
     "require": {
         "php": "^7.4.15 || ^8.0.2",
-        "chrome-php/wrench": "^1.7",
+        "chrome-php/wrench": "^1.8",
         "evenement/evenement": "^3.0.1",
         "monolog/monolog": "^1.27.1 || ^2.8 || ^3.2",
         "psr/log": "^1.1 || ^2.0 || ^3.0",
-        "symfony/filesystem": "^4.4 || ^5.0 || ^6.0 || ^7.0",
+        "symfony/filesystem": "^5.4 || ^6.0 || ^7.0 || ^6.0",
```
jderusse:
```suggestion
        "symfony/filesystem": "^5.4 || ^6.0 || ^7.0 || ^8.0",
```

.github/workflows/tests.yml
```diff
@@ -7,27 +7,35 @@ on:
...
-        symfony: ['4', '5', '6', '7']
+        php: ['7.4', '8.0', '8.1', '8.2', '8.3', '8.4', '8.5']
+        symfony: ['5', '6', '7', '8']
         exclude:
           - php: '7.4'
             symfony: '6'
           - php: '7.4'
             symfony: '7'
+          - php: '7.4'
+            symfony: '8'
```
jderusse, line 22: symfony 8 requires PHP 8.4 🤔 

same for symfony 7.0 that required PHP 8.2

GrahamCampbell: This is an exclude list, not include.

PR #721, opened 2026-01-29
PageNavigation was marked as internal, but is heavely used externally.

MichelYpma:
Promotes PageNavigation to the public API. The class is already widely used in the ecosystem and referred in the documentation, making the @internal flag obsolete and misleading. This change formalizes its status as a supported component of the library.


GrahamCampbell:
Thanks for getting started here. I've replaced this with a more compressive PR: https://github.com/chrome-php/chrome/pull/731.

PR #726, opened 2026-06-18
Allow choosing the box model in `Node::getPosition()` method

aprat84:
When rendering single page documents, this method would be so handy, but it only measures the content-box, so the element padding is omitted.

For this task, and potentially other tasks, it would be useful to be able to chose the box model.

Default to `content`, so it is backwards compatible!

PD: Ideally, a `getBoxModel` method could be added, deprecating the `getPosition`, and return measures for every box model the devtools API returns.

### DIFF
--- a/vendor-bin/phpstan/composer.json
+++ b/vendor-bin/phpstan/composer.json
@@ -1,6 +1,6 @@
 {
     "require": {
-        "php": "^7.4",
+        "php": "^7.4|^8.0",
         "phpstan/phpstan": "2.1.3"
     },
     "config": {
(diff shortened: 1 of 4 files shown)


vendor-bin/phpstan/composer.json
```diff
@@ -1,6 +1,6 @@
 {
     "require": {
-        "php": "^7.4",
+        "php": "^7.4|^8.0",
```
GrahamCampbell, line 3: This change is incorrect. PHPStan **MUST** be run on PHP 7.4 only.

GrahamCampbell (changes requested):

GrahamCampbell:
I've applied some corrections, and merged.

Issue #437, opened 2022-09-17
Custom flag proxy with user pass


lokutus24:
hello,

i did not found any option how can i set custom flag in chrome php for proxy with user pass.
Can somebody help me Please?

enricodias:
There is no native support for this yet, but you can listen for the auth events and authenticate manually. There's an example in the issue #367

alex-bacart:
Unfortunately #367 issue is deleted now. Code from this answer worked for me https://superuser.com/a/1777046/1465413

Issue #158, opened 2020-05-15
Proxy problem


platinumuser:
Hello
How to use a proxy server with authorization login and password?
I indicate such a parameter

`'customFlags' => ['--proxy-server=http://proxy.example.com:8000']`

alex-rsk:
There's no direct possibility in Chromium to use proxies with credentials, you may use proxy forwarding only.  check  [mitmproxy](https://mitmproxy.org/)

TimIgoe:
Its perfectly valid to use proxies with authentication - you just have to catch and respond to the relevant messages in code. (been doing it for many many months)

alex-rsk:
> Its perfectly valid to use proxies with authentication - you just have to catch and respond to the relevant messages in code. (been doing it for many many months)
Firstly, I wrote "**direct**" possibility. Secondly, your way (using Fetch.continueRequest) could cause  segmentation fault sometimes, if you send authorization response to the socket  **after** browser process destroyed.
  Also,  it causes opening empty tabs with 0.0.0.0 (i guess, because browser getting some authorizationRequests **before** you are plugging in to the socket.
I'm for stability. So I chose abovementioned way.

Issue #29, opened 2018-08-06
auth


ph1ex0n:
how to do basic auth with headless chrome?

ph1ex0n:
thanks for answer

ph1ex0n:
can we together find a way to solve this problem?

choval:
@ph1ex0n You could pass it through the URI: 

```
http://user:password@domain/path
```

https://en.wikipedia.org/wiki/Basic_access_authentication#URL_encoding

PR #728, opened 2026-07-05
[1.16] Add `Page::authenticate` to fix proxy issues

### LINKED ISSUE #634: Using a proxy with authorization
Please tell me if it is possible to somehow use a proxy with a login and password?

enricodias:
`Page::setBasicAuthHeader` already exists, but many proxies ignore both  `Authorization` and `Proxy-Authorization` headers and strictly require a challenge-response flow.

This new `Page::authenticate` sends credentials with `Fetch.continueWithAuth` when `Fetch.authRequired` is received. A `CancelAuth` response is used after a failed attempt and an exception is thrown. This should work with any kind of browser auth, not only for proxies.

Resolves #634, #437, #158, #29

### DIFF
--- a/README.md
+++ b/README.md
@@ -682,6 +682,21 @@ if ($cookieBar) {
 }
 ```
 
+### HTTP Authentication
+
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
Thanks for getting this going. I fixed some issues and moved this to https://github.com/chrome-php/chrome/pull/739.

PR #729, opened 2026-07-05
[1.16] Add `AbstractBinaryInput::saveToStream`

### LINKED ISSUE #495: Create PDF to a stream (e.g. using php://memory and php://temp) so it can be passed on the fpassthru
In order to avoid unnecessary memory consumption it would be create if a PDF could be created using a temporary stream like php://temp and directly pass it on to fpassthru without storing the data / decoding it memory (base64). That should avoid unnecessary memory consumption and saveToFile https://github.com/chrome-php/chrome/blob/v1.8.0/src/PageUtils/AbstractBinaryInput.php#L65 already creates a stream, it just stores it to a file and closes it. So adding a corresponding saveToStream or similar should be easily doable.

enricodias:
Similar to `AbstractBinaryInput::saveToFile`, but with a stream. Can be used to pass large PDFs directly to `php://output` without allocating the whole file in memory.

Resolves #495

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,7 +519,11 @@ header('Expires: 0');
 header('Cache-Control: must-revalidate, post-check=0, pre-check=0');
 header('Pragma: public');
 
-echo base64_decode($pdf->getBase64());
+$outputStream = \fopen('php://output', 'w');
+
+$pdf->saveToStream($outputStream);
+
+\fclose($outputStream);
 ```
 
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Thanks for getting started here. I've fixed some issues and moved it to https://github.com/chrome-php/chrome/pull/741.

PR #737, opened 2026-07-05
[1.16] Allow choosing the box model in `Node::getPosition()`

GrahamCampbell:
Node::getPosition() reads only the content box from DOM.getBoxModel, so there is no way to measure an element including its padding, border or margin, which is useful when rendering single page documents among other things. This picks up @aprat84's change from #726, retargeted at 1.16 since it adds new behaviour: getPosition() accepts an optional box model argument defaulting to content, so existing calls are unaffected, and the original commit is included as authored.

A follow-up commit applies some corrections. An invalid box model now throws an InvalidArgumentException, following the precedent of the screenshot format validation, rather than silently returning null, since a null return otherwise means the element could not be measured and a typo should not masquerade as that. The test is rewritten to use a data provider instead of PHP 8 attributes, because #[TestWith] requires PHPUnit 10 and the trailing comma in the parameter list is a parse error on PHP 7.4, so the original test file would have broken the PHP 7.4 and 8.0 builds, which also makes the phpstan-ignore workarounds unnecessary. The PHPStan platform constraint is left untouched as PHPStan must run on PHP 7.4 only, and a changelog entry is included.

Closes #726.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## 1.16.0 (UPCOMING)
 
 * Add `disableJavascript` option
+* Allow choosing the box model in `Node::getPosition()`
 
 
 ## 1.15.0 (2025-12-27)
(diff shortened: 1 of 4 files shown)


aprat84:
Thanks! 🎉

